PR #3, opened 2018-03-05
Optional constructor rounding

omus:
Only round an `AnchoredInterval` when the keyword `round` is set.

### DIFF
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -48,8 +48,8 @@ struct AnchoredInterval{P, T} <: AbstractInterval{T}
     anchor::T
     inclusivity::Inclusivity
 
-    function AnchoredInterval{P, T}(anchor::T, inc::Inclusivity) where {P, T}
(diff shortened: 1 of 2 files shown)


spurll:
The issue that I have with this approach is that it allows the user to pass in a `round=true` kwarg that doesn't actually do something when `P` isn't a `Period`. I think a cleaner solution is to just create a pair of pseudoconstructors `HE` and `HB` that do the rounding. (I've done this in the branch that I'm working on now, but am still working a fix for the display of non-rounded `AnchoredInterval`s.)

omus:
We could also support rounding for non-Period types

spurll:
Yep. The syntax would be slightly different (using the value for `Period` and the type for everything else). I no longer think it's worth implementing for the type itself, though, given that our only use case is for hour ending and we were talking about having a shorter pseudoconstructor anyway.

omus:
Ok, I expect I'll close this. I'll leave it open for now as things change around this.

PR #5, opened 2018-03-06
Create Endpoint struct to make it easier to write Interval comparisons

omus:


omus:
Coverage reduction from me failing to test for `in(a::AbstractInterval{T}, b::AbstractInterval{T}) where T`. Tests for this method will be in another PR. I can either remove this method or we can merge this as is.

PR #6, opened 2018-03-06
Fix rounding issues in `AnchoredInterval`s

### LINKED ISSUE #1: Addition/subtraction for some `AnchoredInterval`s produces unexpected results
Arithmetic involving an `AnchoredInterval{P, T}` when `P isa Period` runs afoul of the rounding code in the constructor, producing counterintuitive results:

```julia
julia> he = HourEnding(now())
HourEnding{DateTime}(2018-02-28T18:00:00, Inclusivity(false, true))

julia> he - Minute(30)
HourEnding{DateTime}(2018-02-28T18:00:00, Inclusivity(false, true))

julia> he + Minute(30)
HourEnding{DateTime}(2018-02-28T19:00:00, Inclusivity(false, true))
```

Rounding will have to be removed from the `AnchoredInterval{P, T}` constructor to resolve this issue, but it is *nice* being able to call `HourEnding(now())` without having to explicitly round `now()` to the nearest hour.

Of the many possible solutions @omus and I have considered, two seem most promising:

1. `HourEnding{T}` remains a type alias for `AnchoredInterval{Hour(-1), T}`, but only the `HourEnding` constructor will perform the rounding operation
    * This means that `HourEnding{DateTime}(now())` will *not* be rounded (as it's just an alias for `AnchoredInterval{Hour(-1), DateTime}`), which may lead to confusion
2. `HourEnding` and `AnchoredInterval` both become subtypes of a new `AbstractAnchoredInterval` (Curt has suggested `SpanInterval` as a replacement name)
    * This doesn't sit particularly well with me because, aside from the rounding that happens in the constructor, the functionality will be identical

Either way, I'll have to fix the way `string(::AnchoredInterval{P, T})` works, because currently it assumes that everything will be rounded when `T <: TimeType`.

spurll:
Remove rounding from `AnchoredInterval{P}` (which was a special case when `P` is a `Period`) and replace it with the pseudoconstructors `HE` and `HB`. This fixes some problems associated arithmetic involving `AnchoredInterval`s.

Also `intersect(AnchoredInterval, AnchoredInterval)` now returns an `AnchoredInterval` instead of an `Interval` (but the tests are broken; they'll be fixed once we change the value type parameter to a field).

Closes #1


spurll:
I'd appreciate your thoughts on this, @omus. (Some of the tests are currently marked as broken, but will be fixed once we get rid of that value type parameter.)

spurll:
Ooof, better fix that coverage before I merge...

src/interval.jl
```diff
@@ -156,6 +156,13 @@ Base.:-(a::Interval{T}, b::T) where T = a + -b
 Base.:-(a::Interval{Char}, b::Integer) = a + -b
 Base.:-(a::Interval{T}, b::Period) where T <: TimeType = a + -b
 
+Base.:-(a::T, b::Interval{T}) where T <: Number = a + -b
+
+function Base.:-(a::Interval{T}) where T <: Number
+    inc = inclusivity(a)
+    Interval{T}(-last(a), -first(a), Inclusivity(last(inc), first(inc)))
+end
```
omus: Consider dropping `<: Number` as its possible other types support unary minus outside of numbers.

src/interval.jl
```diff
@@ -168,10 +175,7 @@ Base.isless(a::T, b::AbstractInterval{T}) where T = a < LeftEndpoint(b)
 ##### SET OPERATIONS #####
 
 function Base.isempty(i::Interval)
-    return (
-        first(i) > last(i) ||   # Shouldn't be possible, but check anyway.
-        (first(i) == last(i) && inclusivity(i) != Inclusivity(true, true))
-    )
+    return first(i) == last(i) && inclusivity(i) != Inclusivity(true, true)
```
omus: Once https://github.com/invenia/Intervals.jl/pull/7 is merged we can change this to:
```julia
LeftEndpoint(i) == RightEndpoint(i)
```

omus: Should I make this change in #7?

src/interval.jl
spurll: Yeah, I wasn't sure, because I definitely want it to fail for `Char` and `TimeType`, but I guess I can just let it fail on the unary minus itself instead of throwing a `MethodError` here.

src/interval.jl
spurll: No, I don't think that will work. You'll still need to verify that at least one of the endpoints is exclusive in order for the interval to be empty:

```
isempty(Interval(10, 10, false, false)) == true
isempty(Interval(10, 10, true, true)) == false    # would be true in your example
isempty(Interval(10, 10, true, false)) == true    # would be false in your example
isempty(Interval(10, 10, false, true)) == true    # would be false in your example
```

src/anchoredinterval.jl
```diff
@@ -149,6 +172,13 @@ Base.:-(a::AnchoredInterval, b) = a + -b
 # Required for StepRange{<:AnchoredInterval}
 Base.:-(a::AnchoredInterval, b::AnchoredInterval) = anchor(a) - anchor(b)
 
+Base.:-(a::T, b::AnchoredInterval{P, T}) where {P, T <: Number} = a + -b
+
+function Base.:-(a::AnchoredInterval{P, T}) where {P, T <: Number}
+    inc = inclusivity(a)
+    AnchoredInterval{-P, T}(-anchor(a), Inclusivity(last(inc), first(inc)))
+end
```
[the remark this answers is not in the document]
spurll: Ah, but we're *not*, right, because this doesn't actually work for `HourEnding` and `HourBeginning`; it's strictly for numerics.

I didn't originally have it doing this, but after some reflection this seemed pretty solidly like the way to do it to me. Because the unary minus flips the interval around the origin point, I believe it should also flip the "direction" of the interval.

src/description.jl
```diff
@@ -0,0 +1,142 @@
...
+end
+
+function description(dt::AbstractDateTime, p::Period, suffix::String)
+    ts = time_string(dt, p)
+
+    # If we would display only HE00, display HE24 for the previous day instead
+    if p isa Hour && ts == "00" && suffix == "E"
+        ts = "24"
+        dt -= Day(1)
+    end
```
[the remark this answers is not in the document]
spurll: Yep. I do not want to generalize it, though. ME60 is silly. (HE24 is silly, too, but it's the industry standard.)

src/description.jl
```diff
@@ -0,0 +1,142 @@
...
+        return "$(prefix(p))$suffix $ds$(ts in ("00", "24") ? "" : " $ts")"
+    end
+end
+
+function time_string(dt::AbstractDateTime, p::Period)
+    ts = @sprintf("%02d:%02d:%02d.%03d", hour(dt), minute(dt), second(dt), millisecond(dt))
+
+    # Trim excess precision from time_string
+    max_replacements = p isa Second ? 1 : p isa Minute ? 2 : 3
+    return replace(ts, Regex("([:.]00+){1,$max_replacements}+\$"), "")
```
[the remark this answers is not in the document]
spurll: I hate how messy it looks, but speed is more important of course. I'll make that change. Thanks.

src/interval.jl
omus: My original comment about the implementation is definitely wrong. As it turns out it would be:
```julia
isempty(i::AbstractInterval) = LeftEndpoint(i) > RightEndpoint(i)
```

src/description.jl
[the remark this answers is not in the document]
spurll: Actually it looks like I need to make this even messier, unfortunately.

src/interval.jl
spurll: I'm not super familiar with the way the endpoint inclusivity checks are done, but so long as the existing tests pass, this should be fine.

omus: Since we're not using the new `==` in #7 we could actually do this implementation here. I'll just make a new PR for this change to keep everything separated

src/anchoredinterval.jl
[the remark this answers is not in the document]
omus: Ah, I missed the `Number`. Yup this is the right way to do this

src/description.jl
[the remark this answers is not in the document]
omus: That is fine. You may still want to take the general implementation in case this needs to change in the future.

src/interval.jl
omus: https://github.com/invenia/Intervals.jl/pull/10

spurll: I'll merge that in next. Thanks.

src/anchoredinterval.jl
```diff
@@ -206,14 +208,23 @@ function Base.intersect(a::AnchoredInterval{P, T}, b::AnchoredInterval{Q, T}) wh
...
+
+function promote_period(p::Period, target_type::Type{<:Period}=Day)
+    if !isa(p, target_type)
+        P, max_val = coarserperiod(typeof(p))
+        if (value(p) % max_val == 0) && (max_val > 1)
+            p = promote_period(convert(P, p), target_type)
+        end
+    end
+    return p
+end
```
omus: If you define:
```julia
promote_period(p::P, target_type::Type{P}) where P <: Period = p
```
You can remove the `isa`

src/anchoredinterval.jl
```diff
@@ -206,14 +208,23 @@ function Base.intersect(a::AnchoredInterval{P, T}, b::AnchoredInterval{Q, T}) wh
...
 
-    # TODO: P is always going to be milliseconds... :(
-    # Update this once P is replaced by a span element
-
     return AnchoredInterval{new_P, T}(anchor, inclusivity(interval))
 end
+
+##### UTILITIES #####
+
+function promote_period(p::Period, target_type::Type{<:Period}=Day)
```
omus: Maybe we should reuse the `canonicalize` name:
```julia
julia> Base.Dates.canonicalize(Dates.CompoundPeriod(Hour(24)))
1 day
```
Probably something like:
```julia
function canonicalize(target_type::Type{<:Period}, p::Period)
    ...
end
```

omus: It's also weird to call this "promote" as usually those functions work in the opposite direction:
```julia
julia> promote_type(Hour, Day)
Base.Dates.Hour
```

omus:
Do you want this merged or can it be squashed?

PR #7, opened 2018-03-06
Endpoint equality

omus:
Verifies that endpoint `==` works. ~I need this~ Used in an upcoming PR.

### DIFF
--- a/src/endpoint.jl
+++ b/src/endpoint.jl
@@ -21,6 +21,26 @@ RightEndpoint(ep::T, included::Bool) where T = RightEndpoint{T}(ep, included)
 LeftEndpoint(i::AbstractInterval{T}) where T = LeftEndpoint{T}(first(i), first(inclusivity(i)))
 RightEndpoint(i::AbstractInterval{T}) where T = RightEndpoint{T}(last(i), last(inclusivity(i)))
 
+
+"""
+    ==(a::LeftEndpoint{T}, b::RightEndpoint{T}) where T -> Bool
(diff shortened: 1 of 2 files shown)


src/endpoint.jl
```diff
@@ -21,6 +21,14 @@ RightEndpoint(ep::T, included::Bool) where T = RightEndpoint{T}(ep, included)
 LeftEndpoint(i::AbstractInterval{T}) where T = LeftEndpoint{T}(first(i), first(inclusivity(i)))
 RightEndpoint(i::AbstractInterval{T}) where T = RightEndpoint{T}(last(i), last(inclusivity(i)))
 
+function Base.:(==)(a::LeftEndpoint{T}, b::RightEndpoint{T}) where T
+    a.endpoint == b.endpoint && a.included && b.included
+end
```
spurll: I think I understand what you're getting at here, but it could definitely use some documentation, because I think most people looking at this code are going to spend some time wondering why this isn't

```julia
a.endpoint == b.endpoint && a.included == b.included
```

spurll: That minor point aside, this is 👍.

omus:
Fair points. I'll add some documentation to try and make things clearer.

spurll:
Thanks. After this one goes in I think I'm going to tag a v0.1.0 release; let me know if you think any of the other PRs should get in before then.

omus:
I've added the documentation as a docstring. Maybe I should do a comment instead as the docstring is exported even though these types are not.

PR #8, opened 2018-03-06
Interval comparison tests

omus:
A new test suite which performs a variety of comparisons on the various possible kinds of intervals. I found this to be rather useful for catching corner cases as I modified the code.

I've also modified `isless` and `<` to be different. The behaviour of `isless` which is used for sorting just checks the left-endpoint. When the left-endpoints are equal the two intervals are considered equal for sorting and are produced in the order in which they appear.

### DIFF
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -185,7 +185,10 @@ end
 
 # Required for min/max of AnchoredInterval{LaxZonedDateTime} when the anchor is AMB or DNE
 function Base.:<(a::AnchoredInterval{P, T}, b::AnchoredInterval{P, T}) where {P, T}
-    return anchor(a) < anchor(b)
+    return (
+        anchor(a) < anchor(b) ||
+        (anchor(a) == anchor(b) && first(inclusivity(a)) && !first(inclusivity(b)))
+    )
 end
 
 ##### RANGE #####
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -125,6 +125,31 @@ function Base.convert(::Type{Interval{T}}, interval::AnchoredInterval{P, T}) whe
     return Interval{T}(first(interval), last(interval), inclusivity(interval))
 end
 
+# Conversion methods which currently aren't needed but could prove useful. Commented out
+# since these are untested.
+
+#=
+function Base.convert(::Type{AnchoredInterval{P, T}}, interval::Interval{T}) where {P, T}
+    @assert abs(P) == span(interval)
+    anchor = P < zero(P) ? last(interval) : first(interval)
+    AnchoredInterval{P, T}(last(interval), inclusivity(interval))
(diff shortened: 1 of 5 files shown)


src/anchoredinterval.jl
```diff
@@ -1,3 +1,7 @@
+abstract type Anchor end
+struct Ending <: Anchor end
+struct Beginning <: Anchor end
```
omus: I think I can re-use `Direction` here. I wrote this code before endpoints were introduced.

src/anchoredinterval.jl
```diff
@@ -100,6 +104,30 @@ function Base.convert(::Type{Interval{T}}, interval::AnchoredInterval{P, T}) whe
...
+    AnchoredInterval{P, T}(anchor, inclusivity(interval))
+end
+
+function Base.convert(::Type{AnchoredInterval{Ending}}, interval::Interval{T}) where {T}
+    AnchoredInterval{-span(interval), T}(last(interval), inclusivity(interval))
+end
+
+function Base.convert(::Type{AnchoredInterval{Beginning}}, interval::Interval{T}) where {T}
+    AnchoredInterval{span(interval), T}(first(interval), inclusivity(interval))
+end
```
omus: I need to do some strange things to say I want an `AnchoredInterval{<0}`...

src/anchoredinterval.jl
```diff
@@ -100,6 +104,30 @@ function Base.convert(::Type{Interval{T}}, interval::AnchoredInterval{P, T}) whe
...
+    @assert abs(P) == span(interval)
+    anchor = P < zero(P) ? last(interval) : first(interval)
+    AnchoredInterval{P, T}(last(interval), inclusivity(interval))
+end
+
+function Base.convert(::Type{AnchoredInterval{P}}, interval::Interval{T}) where {P, T}
+    @assert abs(P) == span(interval)
+    anchor = P < zero(P) ? last(interval) : first(interval)
+    AnchoredInterval{P, T}(anchor, inclusivity(interval))
+end
```
omus: I'll comment these out as they are currently unused and untested

omus: Done

omus:
AppVeyor 32-bit failure on Julia 0.6 is unrelated (WinRPM download failure). [Code coverage looks good](https://codecov.io/gh/invenia/Intervals.jl/pull/8?src=pr&el=h1)

omus:
Re-running AppVeyor fixed the CI

spurll:
Unless there's something critical in here, I'm planning to hold off on reviewing this stuff for now.

spurll:
I'd be happy to take a look at this next week, but it has some conflicts with the changes we made in #13. If you have time to resolve those let me know, but I don't see this as a particularly pressing PR.

omus:
I'll fix the conflicts

PR #10, opened 2018-03-07
General isempty implementation

omus:


omus:
Rebased to deal with conflicts from #6

PR #12, opened 2018-03-08
WIP: Move away from using span value as a type parameter

omus:
Moving away from using the span value as a type parameter as this does not allow us to check properties of the type like "ending" (span is < 0) or "beginning" (span is > 0).


omus:
The current commit switches `AnchorInterval{P, T}` to `AnchorInterval{T, P}` in preparation for additional commits which end up changing `AnchorInterval` to its final form of `AnchorInterval{T, S, E}` where `S` is the type of the span and `E` is an instance of `Direction` (either `Ending` or `Beginning`).

omus:
Currently the `span` can field can be positive/negative which is strange as the `span` accessor only returns  a positive value. I believe the code can be simplified if we restrict the span parameter passed into `AnchoredInterval` to be always positive.

omus:
Note that with these changes since the value of the span is no longer part of the type things like `HourEnding` have spans of `Hour` but not necessarily one hour:

```julia
julia> HourEnding(now(), Hour(3))
HourEnding{DateTime}(2018-03-08T14:44:00.085, 3 hours, Inclusivity(false, true))
```
This does have benefits as it makes our intersect function always return the same type:
```julia
julia> dt = now()
2018-03-08T14:44:46.966

julia> intersect(IntervalEnding(dt, Hour(3)), HourEnding(dt - Hour(1), Hour(3)))
HourEnding{DateTime}(2018-03-08T13:44:46.966, 2 hours, Inclusivity(false, true))

julia> ans isa IntervalEnding{DateTime, Hour}
true
```

spurll:
I am eying this PR with a mix of excitement and trepidation. I look forward to reviewing this at a later date.

omus:
Fair enough. I would like to get this merge next week sometime so let me know if you want to go through it together.

spurll:
That would be good. I'm not sure I'll have time next week, but I'll let you know.

PR #13, opened 2018-03-08
Replace existing `isless` with `<` and `≪` (`less_than_disjoint`)

### LINKED ISSUE #14: Use `<` (and `≪`) instead of `isless`
In consulting with @omus and @iamed2, I've determined that we should have implemented `<` instead of `isless`. The existing implementation of `isless` also assumes that an interval is only less than another interval if they are disjoint (which makes it difficult if you want to, say, sort things or do simple `min` and `max`).

So:
* `isless` should be stricken
* `<` will be defined such that it doesn't require the intervals to be disjoint
* `≪` (`muchless`? `ismuchless`? `disjoint_less_than`?) will be defined as `a < b && isempty(a ∩ b)` (`≫` will also have to be defined)

spurll:
Current implementation of `isless(a, b)` assumes that you need `a` and `b` to be disjoint, which is an unwarranted assumption. Also, as per #14, `isless` is the wrong method to define, because we don't have an absolute canonical ordering for intervals. So that is being resolved.

From the new documentation:

>The standard `<` and `>` operators simply compare the leftmost endpoint of the intervals,
and are used for things like `sort`, `min`, `max`, etc.
>
>The `≪` and `≫` operators (the Unicode symbols for "much less than" and "much greater than",
accessible from the REPL with `\ll` and `\gg`, respectively) are used in this context to
mean "less/greater than and disjoint"; they will verify that there is no overlap between
the intervals.

```julia
julia> 0..10 < 10..20
true

julia> 0..10 ≪ 10..20
false

julia> 0..10 ≪ 11..20
true
```

Closes #14, #16.

--

Additionally, this resolves another bug that results from the disjoint assumption when dealing with `AnchoredInterval`s. For most purposes, the default less than comparison used for `AbstractInterval`s is fine, but because it compares left and right endpoints, and in some rare cases (e.g., if you have an `AnchoredInterval` of `LaxZonedDateTime`s, one of which is ambiguous/nonexistent) it's not possible to calculate the second endpoint using the anchor, this can result in some strange behaviour:

```julia
julia> using Intervals, TimeZones, LaxZonedDateTimes

julia> start = LaxZonedDateTime(DateTime(2016, 11, 10, 1, 45), tz"America/Winnipeg")
2016-11-10T01:45:00-06:00

julia> finish = LaxZonedDateTime(DateTime(2016, 11, 6, 1, 45), tz"America/Winnipeg")
2016-11-06T01:45:00-AMB

julia> collect(HourEnding(start):Day(-2):HourEnding(finish))
3-element Array{HourEnding{LaxZonedDateTimes.LaxZonedDateTime},1}:
    (2016-11-10 HE01:45:00]
 #undef
 #undef
```

This PR also resolves that issue.

Related: It's probably time to open-source LaxZonedDateTimes.jl.


src/Intervals.jl
```diff
@@ -28,6 +28,10 @@ export AbstractInterval,
        isclosed,
        anchor,
        span,
-       ..
+       disjoint_less_than,
+       disjoint_greater_than,
```
omus: If we expand the amount of greater/less than functions I could possibly see us going with `less_than_disjoint` for easy discoverability. I'm fine with leaving this and just doing deprecations if we decide to switch things.

src/anchoredinterval.jl
```diff
@@ -181,6 +181,13 @@ function Base.:-(a::AnchoredInterval{P, T}) where {P, T <: Number}
     AnchoredInterval{-P, T}(-anchor(a), Inclusivity(last(inc), first(inc)))
 end
 
+##### EQUALITY #####
+
+# Required for min/max of AnchoredInterval{LaxZonedDateTime} when the anchor is ambiguous
```
omus: Ambiguous or non-existent.

src/interval.jl
```diff
@@ -150,12 +150,50 @@ end
...
 
-Base.isless(a::AbstractInterval{T}, b::T) where T = RightEndpoint(a) < b
-Base.isless(a::T, b::AbstractInterval{T}) where T = a < LeftEndpoint(b)
+disjoint_greater_than(a, b) = disjoint_less_than(b, a)
+
+"""
+    ≪(a::AbstractInterval, b::AbstractInterval) -> Bool
+    disjoint_less_than(a::AbstractInterval, b::AbstractInterval) -> Bool
+
+Less-than-and-disjoint comparison operator.
```
omus: Should probably be more descriptive here. For example:

> Check that the interval `a` is less than the interval `b` and that they are disjoint (do not overlap).

test/anchoredinterval.jl
```diff
@@ -289,39 +289,48 @@ using Intervals: canonicalize
...
-        @test !isless(diff_inc, diff_inc + Hour(1))     # Overlap for an instant
+        # Overlap for an instant, so not disjoint
+        @test he < hb
+        @test !(hb < he)
+        @test !(he > hb)
+        @test hb > he
+        @test !(he ≪ hb)
+        @test !(hb ≪ he)
+        @test !(he ≫ hb)
+        @test !(hb ≫ he)
```
omus: Just going to mention that `≪̸` (`\NotLessLess`) and `≫̸` (`\NotGreaterGreater`) exist.

src/interval.jl
```diff
@@ -150,12 +150,50 @@ end
...
+Base.:<(a::AbstractInterval{T}, b::T) where T = RightEndpoint(a) < b
+Base.:<(a::T, b::AbstractInterval{T}) where T = a < LeftEndpoint(b)
+
+function Base.:<(a::AbstractInterval{T}, b::AbstractInterval{T}) where T
+    return LeftEndpoint(a) < LeftEndpoint(b)
+end
+
+function disjoint_less_than(a::AbstractInterval{T}, b::AbstractInterval{T}) where T
     return RightEndpoint(a) < LeftEndpoint(b)
 end
```
omus: Should we implement a version of this for scalars and intervals?

src/Intervals.jl
spurll: Good point. I'll make this change now.

test/anchoredinterval.jl
spurll: Unfortunately there doesn't appear to be a not-much-less-than.

src/interval.jl
spurll: I went back and forth about that a fair bit. `scalar < interval` is always going to mandate disjoint, while `interval < scalar` can have a meaningful distinction here.

Probably makes sense to implement both.

test/anchoredinterval.jl
spurll: Whoops, misread that message because `≪̸` and `≮` are basically indistinguishable on my monitor.

PR #17, opened 2018-03-09
Replace `in` with `issubset`

### LINKED ISSUE #15: Replace `in(::AbstractInterval, ::AbstractInterval)` with `issubset(::AbstractInterval, ::AbstractInterval)`
Because `in` is the wrong function to use in this case.

spurll:
Replace `in(::AbstractInterval, ::AbstractInterval)` with `issubset(::AbstractInterval, ::AbstractInterval)` and define/export `⊆`, `⊇`, `⊈`, and `⊉`

Closes #15

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -182,6 +182,7 @@ true
 ```
 """
 ≪(a, b) = less_than_disjoint(a, b)
+# ≪̸(a, b) = !≪(a, b)
 
 """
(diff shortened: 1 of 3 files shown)


src/interval.jl
```diff
@@ -1,3 +1,5 @@
+
+
```
omus: On purpose?

PR #18, opened 2018-03-09
Fix `<` for `AnchoredInterval`

omus:
Fixes a bug introduced in #13. I only caught this because of the tests in #8. This fix is also included in that PR.

### DIFF
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -185,7 +185,10 @@ end
 
 # Required for min/max of AnchoredInterval{LaxZonedDateTime} when the anchor is AMB or DNE
 function Base.:<(a::AnchoredInterval{P, T}, b::AnchoredInterval{P, T}) where {P, T}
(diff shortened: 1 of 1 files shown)


src/anchoredinterval.jl
```diff
@@ -185,7 +185,10 @@ end
 
 # Required for min/max of AnchoredInterval{LaxZonedDateTime} when the anchor is AMB or DNE
 function Base.:<(a::AnchoredInterval{P, T}, b::AnchoredInterval{P, T}) where {P, T}
-    return anchor(a) < anchor(b)
+    return (
+        anchor(a) < anchor(b) ||
+        (anchor(a) == anchor(b) && first(inclusivity(a)) && !first(inclusivity(b)))
+    )
```
omus: Aside: we also need to enforce that the spans are the same. Since currently `P` is the value of the span this is being enforced by the method itself.

PR #22, opened 2018-05-01
Add astimezone(::Interval{ZonedDateTime}, ::TimeZone)

spurll:
Allows us to easily convert `Interval`s and `AnchoredInterval`s from one time zone to another.


spurll:
Got one more fix coming down the pipeline for #23. I was planning to make this a separate PR, but I can bundle it in. What do you think @iamed2?

iamed2:
Might as well make separate, but merge at your leisure :)

PR #25, opened 2018-05-02
Replace `<` with `isless`

### LINKED ISSUE #19: Missing `isless` means that `sort` fails
...unless you specify `lt=<`.

spurll:
Allows `sort`ing of `AbstractInterval`s.

Closes #19

### DIFF
--- a/test/anchoredinterval.jl
+++ b/test/anchoredinterval.jl
@@ -301,6 +301,8 @@ using Intervals: canonicalize
         @test hash(he) != hash(diff_inc)
 
         # Overlap for an instant, so not disjoint
+        @test isless(he, hb)
+        @test !isless(hb, he)
         @test he < hb
         @test !(hb < he)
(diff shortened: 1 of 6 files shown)


test/anchoredinterval.jl
```diff
@@ -346,6 +350,16 @@ using Intervals: canonicalize
...
     end
 
+    @testset "sort" begin
+        hb1 = HourBeginning(dt)
+        he1 = HourEnding(dt)
+        he2 = HourEnding(dt + Hour(1))
+        he3 = HourEnding(dt + Hour(2))
+
+        @test sort([hb1, he1, he2, he3]) == [he1, hb1, he2, he3]
+        @test sort([hb1, he1, he2, he3]; rev=true) == [he3, he2, hb1, he1]
```
omus: Doesn't honestly matter but these could be `sort!`

PR #27, opened 2018-05-03
Improve docs for equality and DataFrames

spurll:
Adds documentation detailing some special considerations to be made when comparing `Interval`s with `AnchoredInterval`s.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -123,6 +123,30 @@ julia> ZonedDateTime(he)
 
 ### Comparisons
 
+#### Equality
+
+Two `AbstractInterval`s are considered equal if they have identical left and right
+endpoints (taking `Inclusivity` into account):
+
+```julia
+julia> a = Interval(DateTime(2013, 2, 13), DateTime(2013, 2, 13, 1), true, false)
+Interval{DateTime}(2013-02-13T00:00:00, 2013-02-13T01:00:00, Inclusivity(true, false))
+
+julia> b = Interval(DateTime(2013, 2, 13), DateTime(2013, 2, 13, 1), false, true)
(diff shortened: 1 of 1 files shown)


docs/src/index.md
```diff
@@ -123,6 +123,30 @@ julia> ZonedDateTime(he)
 
 ### Comparisons
 
+#### Equality
+
+Two `AbstractInterval`s are considered equal if they have identical left and right
+endpoints (taking `Inclusivity` into account):
```
omus: It would be really great to have docstrings for `LeftEndpoint` and `RightEndpoint`

docs/src/index.md
```diff
@@ -146,17 +170,144 @@ julia> 0..10 ≪ 11..20
...
+│ 7   │ (2016-11-16 HE07] │ 7    │ 'g' │
+│ 8   │ (2016-11-16 HE08] │ 8    │ 'h' │
+│ 9   │ (2016-11-16 HE09] │ 9    │ 'i' │
+│ 10  │ (2016-11-16 HE10] │ 10   │ 'j' │
+│ 11  │ (2016-11-16 HE11] │ 11   │ 'k' │
+│ 12  │ (2016-11-16 HE12] │ 12   │ 'l' │
+
+julia> join(df2, df3; on=:time)
+0×3 DataFrames.DataFrame
+
```
omus: Extra whitespace? I guess this is actually what is produced.

spurll: Correct! It looked weird when I removed it, so I left it in.

docs/src/index.md
omus: Ya, that guy is a slacker ;). I'll open an issue for this as I don't think I'll have the time at the moment: https://github.com/invenia/Intervals.jl/issues/28

PR #32, opened 2018-05-19
Use AutoHashEquals

rofinn:
In particular, `intersect` and `==` will sometimes not work if this isn't used.


rofinn:
N/m looks like it was just an issue with the endpoint equality.

PR #34, opened 2018-05-30
Adds merge to compress overlapping intervals

morris25:


src/interval.jl
```diff
@@ -74,6 +74,44 @@ Interval(f, l, x::Bool, y::Bool) = Interval(f, l, Inclusivity(x, y))
...
+"""
+    merge(intervals::T) where T <: AbstractVector{<:AbstractInterval}
+
+Flattens a vector of overlapping intervals into smaller vector containing only
+non-overlapping intervals.
+"""
+function merge(intervals::AbstractVector{<:AbstractInterval{T}}) where T
+    length(intervals) < 2 && return intervals
+
+    sort!(intervals)
```
ararslan: Note that this will mutate the argument passed to `merge`, so by convention the function should be named `merge!`. If it's not the intent to mutate the input, I recommend using `sort` instead of `sort!`, which will make a copy.

It may be worthwhile to go all-in on the in-place approach and overwrite the input completely during the computation and return the mutated original array. Then the out-of-place version could be defined as `merge(intervals) = merge!(copy(intervals))`. Just a suggestion.

src/interval.jl
```diff
@@ -74,6 +74,44 @@ Interval(f, l, x::Bool, y::Bool) = Interval(f, l, Inclusivity(x, y))
 Interval(f::T, l::T) where T = Interval(f, l, Inclusivity(true, true))
 (..)(f::T, l::T) where T = Interval(f, l)
 
+"""
+    merge(intervals::T) where T <: AbstractVector{<:AbstractInterval}
+
+Flattens a vector of overlapping intervals into smaller vector containing only
+non-overlapping intervals.
+"""
+function merge(intervals::AbstractVector{<:AbstractInterval{T}}) where T
```
omus: Could just be:
```julia
function merge(intervals::AbstractVector{<:AbstractInterval})
    ...
end
```

src/interval.jl
```diff
@@ -74,6 +74,44 @@ Interval(f, l, x::Bool, y::Bool) = Interval(f, l, Inclusivity(x, y))
...
+        # If the two intervals intersect then we expand the previous interval to include the
+        # current one.
+        else
+            left = min(LeftEndpoint(prev), LeftEndpoint(curr))
+            right = max(RightEndpoint(prev), RightEndpoint(curr))
+            prev = Interval(
+                left.endpoint,
+                right.endpoint,
+                Inclusivity(left.included, right.included)
+            )
```
omus: I think this could possibly be the beginnings of the `union(::AbstractInterval, ::AbstractInterval)` method. However when the two intervals are disjoint we would need to return both intervals. Probably this would result in us returning a set or an vector which probably means additional changes to the already existing `intersect` function.

This leads me to believe that this method should be `union(intervals::AbstractVector{<:AbstractInterval{T}})`

omus: Which also means this method should be moved to the set operations section below

morris25: `merge` is dead, long live `union`

src/interval.jl
```diff
@@ -222,15 +222,59 @@ end
...
+            deleteat!(intervals, i)
+        end
+    end
+
+    return intervals
+end
```
rofinn: 2 Suggestions:

1. Rather than calling `length` on every loop, just update a variable `n` when we delete an entry.
2. Turn the Interval merging into a constructor.

Example)
```
function Interval(a::Interval, b::Interval)
    isempty(intersect(a, b)) && throw(ArgumentError(...))

    left = min(LeftEndpoint(prev), LeftEndpoint(curr))
    right = max(RightEndpoint(prev), RightEndpoint(curr))
    
    return Interval(
        left.endpoint,
        right.endpoint,
        Inclusivity(left.included, right.included)
    )
end

function Base.merge!(intervals::AbstractVector{<:AbstractInterval{T}}) where T
    sort!(intervals)
    n = length(intervals)

    i = 2
    while i <= n
        prev = intervals[i - 1]
        curr = intervals[i]

        # If the current and previous intervals don't intersect then move along
        if isempty(intersect(prev, curr))
            i = i + 1

        # If the two intervals intersect then we absorb the current interval into
        # the previous one.
        else
            intervals[i - 1] = Interval(prev, curr)
            n = length(deleteat!(intervals, i))
        end
    end

    return intervals
end
```

morris25: The constructor can already take 2 `Intervals` so it looks like we're getting `merge` after all

rofinn: Lol, fair enough. I'd also be fine with overloading `+`, as we're already using it to extend the endpoints of an interval.

https://github.com/invenia/Intervals.jl/blob/master/src/interval.jl#L141

We could just have a dispatch for `+(::Interval, ::Interval)`.

omus: I don't think `+(::Interval, ::Interval)` makes much sense. Currently addition is used to shift a range on the number/date line. I think this would end up being `union(::Interval, ::Interval)` which would need to return a `Set{Interval}` to handle the case where the intervals are disjoint. For now I thing that is out of the scope of this PR.

omus:
There are some tests in [comparison.jl](https://github.com/invenia/Intervals.jl/blob/master/test/comparisons.jl#L41) which should be enabled. They were originally thought to be for `union` but `merge` would be appropriate for them. @morris25 you should update those tests and enable them.

src/interval.jl
```diff
@@ -222,15 +234,55 @@ end
...
+        # If the current and previous intervals don't intersect then move along
+        if isempty(intersect(prev, curr))
+            i = i + 1
+
+        # If the two intervals intersect then we absorb the current interval into
+        # the previous one.
+        else
+            intervals[i - 1] = merge(prev, curr)
+            deleteat!(intervals, i)
+            n = n-1
```
omus: Spaces around `-`

ararslan: You can also write this as `n -= 1`

test/comparisons.jl
```diff
@@ -96,7 +98,8 @@ const INTERVAL_TYPES = [Interval, AnchoredInterval{Ending}, AnchoredInterval{Beg
         @test !issubset(later, earlier)
 
         @test isempty(intersect(earlier, later))
-        # @test union(a, b) == Interval(1, 5, false, true)
+        @test_throws ArgumentError merge(earlier, later)
+        @test union([earlier, later]) == [earlier, later]
```
omus: These should be combined into a single interval as there is nothing that can exist between these two intervals.

test/comparisons.jl
```diff
@@ -125,7 +128,8 @@ const INTERVAL_TYPES = [Interval, AnchoredInterval{Ending}, AnchoredInterval{Beg
         @test !issubset(later, earlier)
 
         @test isempty(intersect(earlier, later))
-        # @test union(a, b) == Interval(1, 5, true, false)
+        @test_throws ArgumentError merge(earlier, later)
+        @test union([earlier, later]) == [earlier, later]
```
omus: Also should result in a single merged interval

src/interval.jl
```diff
@@ -222,15 +252,57 @@ end
...
 
+# There is power in a union.
+"""
+    union(intervals::AbstractVector{<:AbstractInterval})
+
+Flattens a vector of overlapping intervals into a new, smaller vector containing only
+non-overlapping intervals.
+"""
+function Base.union(intervals::AbstractVector{<:AbstractInterval})
+
```
omus: Extra whitespace

src/interval.jl
```diff
@@ -84,6 +84,36 @@ end
...
+    return left <= right
+end
+
+function contiguous(a::AbstractInterval, b::AbstractInterval)
+    left = max(LeftEndpoint(a), LeftEndpoint(b))
+    right = min(RightEndpoint(a), RightEndpoint(b))
+
+    return right.endpoint == left.endpoint && (left.included || right.included)
+end
```
omus: Both `overlaps` and `contiguous` should be added to the "tests/comparisons.jl" gauntlet

test/comparisons.jl
```diff
@@ -160,6 +168,8 @@ const INTERVAL_TYPES = [Interval, AnchoredInterval{Ending}, AnchoredInterval{Beg
         @test intersect(earlier, later) == Interval(3, 3, true, true)
         @test merge(earlier, later) == Interval(1, 5, true, true)
         @test union([earlier, later]) == [Interval(1, 5, true, true)]
+        @test overlaps(earlier, later)
+        @test contiguous(earlier, later)
```
omus: Hmmm, I think `contiguous` should be `false` here

omus: This brings a new test to mind. For intervals using discrete units like integers the intervals `[1,2]` and `[3,4]` should be contiguous. I'll make that a separate issue.

omus: Issue I mentioned: https://github.com/invenia/Intervals.jl/issues/36

src/interval.jl
```diff
@@ -84,6 +84,38 @@ end
...
+    return left <= right
+end
+
+function contiguous(a::AbstractInterval, b::AbstractInterval)
+    left = max(LeftEndpoint(a), LeftEndpoint(b))
+    right = min(RightEndpoint(a), RightEndpoint(b))
+
+    return right.endpoint == left.endpoint &&
+        (left.included || right.included) &&
+        left.included != right.included
```
omus: Could simply be `right.endpoint == left.endpoint && left.included != right.included`

PR #40, opened 2018-06-06
Refactoring of #34 and #37

omus:


src/interval.jl
```diff
@@ -248,12 +217,26 @@ true
 Base.isempty(i::AbstractInterval) = LeftEndpoint(i) > RightEndpoint(i)
 Base.in(a::T, b::AbstractInterval{T}) where T = !(a ≫ b || a ≪ b)
```
omus: The `where` is needed here otherwise it appears that `in` works for comparing intervals

omus:
AppVeyor failure is currently expected.

PR #42, opened 2018-07-14
Add superset function

omus:
- Add `superset` function that creates an interval which encompasses all of the intervals within an array
- ~Rename interval check functions to start with "is" (with deprecations)~
- Add `Interval` constructor which takes `Endpoint`s

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -84,6 +84,13 @@ Interval(f, l, inc...) = Interval(promote(f, l)..., inc...)
 Interval(interval::AbstractInterval) = convert(Interval, interval)
 Interval{T}(interval::AbstractInterval) where T = convert(Interval{T}, interval)
 
+# Endpoint constructors
+function Interval{T}(left::LeftEndpoint{T}, right::RightEndpoint{T}) where T
+    Interval{T}(left.endpoint, right.endpoint, left.included, right.included)
+end
+
+Interval(left::LeftEndpoint{T}, right::RightEndpoint{T}) where T = Interval{T}(left, right)
+
 # Empty Intervals
 Interval{T}() where T = Interval{T}(zero(T), zero(T), Inclusivity(false, false))
 Interval{T}() where T <: TimeType = Interval{T}(T(0), T(0), Inclusivity(false, false))
@@ -250,7 +257,7 @@ function Base.intersect(a::AbstractInterval{T}, b::AbstractInterval{T}) where T
     left = max(LeftEndpoint(a), LeftEndpoint(b))
     right = min(RightEndpoint(a), RightEndpoint(b))
 
-    return Interval{T}(left.endpoint, right.endpoint, left.included, right.included)
+    return Interval{T}(left, right)
 end
 
 # There is power in a union.
@@ -302,11 +309,7 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
 
     left = min(LeftEndpoint(a), LeftEndpoint(b))
     right = max(RightEndpoint(a), RightEndpoint(b))
-    return Interval(
-        left.endpoint,
-        right.endpoint,
-        Inclusivity(left.included, right.included)
-    )
+    return Interval(left, right)
 end
 
 ##### TIME ZONES #####
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -302,6 +302,18 @@ function Base.union!(intervals::Union{AbstractVector{<:Interval}, AbstractVector
     return intervals
 end
 
+"""
+    superset(intervals::AbstractArray{<:AbstractInterval}) -> Interval
+
+Create the smallest single interval which encompasses all of the provided intervals.
+"""
+function superset(intervals::AbstractArray{<:AbstractInterval})
+    left = minimum(LeftEndpoint.(intervals))
+    right = maximum(RightEndpoint.(intervals))
+
+    return Interval(left, right)
+end
+
 function Base.merge(a::AbstractInterval, b::AbstractInterval)
     if !overlaps(a, b) && !contiguous(a, b)
         throw(ArgumentError("$a and $b are neither overlapping or contiguous."))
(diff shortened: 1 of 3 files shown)


src/interval.jl
```diff
@@ -288,18 +295,26 @@ function Base.union!(intervals::Union{AbstractVector{<:Interval}, AbstractVector
     return intervals
 end
 
+"""
+    superset(intervals::AbstractArray{<:AbstractInterval}) -> Interval
+
+Create a single interval which encompasses all of the provided intervals.
```
ararslan: Perhaps it would be worthwhile to mention that this is the smallest such interval?

src/interval.jl
```diff
@@ -288,18 +295,26 @@ function Base.union!(intervals::Union{AbstractVector{<:Interval}, AbstractVector
...
+"""
+function superset(intervals::AbstractArray{<:AbstractInterval})
+    left = minimum(Intervals.LeftEndpoint.(intervals))
```
ararslan: It should be more efficient to only make one pass over the array by providing the function as an argument to `maximum` and `minimum`, e.g.
```julia
left = minimum(LeftEndpoint, intervals)
right = maximum(RightEndpoint, intervals)
```
(Also note that the module prefix shouldn't be necessary within the module.)

src/deprecated.jl
```diff
@@ -0,0 +1,10 @@
+using Base: @deprecate
+
+# BEGIN Intervals 0.1 deprecations
+
+@deprecate less_than_disjoint(a, b) is_less_than_disjoint(a, b) true
```
ararslan: Would it make sense to reuse some Base terminology here and call it like `isless_disjoint`?

src/interval.jl
omus: Thanks. I was developing this in the REPL which is why I have the module prefix

omus: Thanks. I was developing this in the REPL which is why I have the module prefix

src/deprecated.jl
omus: That is better. Do you think that `greater_than_disjoint(a, b)` should be deprecated to `isless_disjoint(b, a)`?

src/interval.jl
ararslan: That's... really weird. 🤔

src/deprecated.jl
ararslan: Yeah, I think that seems reasonable.

ararslan:
Looks like that was https://github.com/JuliaLang/julia/pull/27302.

omus:
Rebased on top of #43 which should address the 0.7 issues

omus:
32-bit issues on AppVeyor with Coverage. Intervals test all pass.

src/interval.jl
```diff
@@ -167,24 +174,21 @@ end
...
+
+function isless_disjoint(a::AbstractInterval, b::AbstractInterval)
     return RightEndpoint(a) < LeftEndpoint(b)
 end
 
-greater_than_disjoint(a, b) = less_than_disjoint(b, a)
-
 """
     ≪(a::AbstractInterval, b::AbstractInterval) -> Bool
-    less_than_disjoint(a::AbstractInterval, b::AbstractInterval) -> Bool
```
ararslan: Perhaps just change the reference here to `isless_disjoint`?

omus: I ended up removing the reference here as there isn't a `isgreater_disjoint` for the other docstring. Additionally, this specific docstring will only appear with `help?> ≪` and not with `help?> isless_disjoint`

src/interval.jl
```diff
@@ -228,26 +231,26 @@ end
 Base.:⊈(a::AbstractInterval, b::AbstractInterval) = !issubset(a, b)
 Base.:⊉(a::AbstractInterval, b::AbstractInterval) = !issubset(b, a)
 
-function overlaps(a::AbstractInterval, b::AbstractInterval)
+function isoverlapping(a::AbstractInterval, b::AbstractInterval)
```
iamed2: I think `overlaps` is easier to read and has precedent with `contains` as a predicate name.

src/interval.jl
```diff
@@ -228,26 +231,26 @@ end
...
     left = max(LeftEndpoint(a), LeftEndpoint(b))
     right = min(RightEndpoint(a), RightEndpoint(b))
 
     return left <= right
 end
 
-function contiguous(a::AbstractInterval, b::AbstractInterval)
+function iscontiguous(a::AbstractInterval, b::AbstractInterval)
```
iamed2: I think `is` implies a predicate with one value. I think `contiguous` is fine but I don't have an issue with changing this. If you do change it, however, it should be `arecontiguous`.

src/interval.jl
ararslan: Though `contains` is deprecated in 0.7 in favor of `occursin`.

iamed2: Same deal with plural predicate though

ararslan: I think Jeff's rule of thumb for 2-arg predicates is that it reads like English if the function is put between the arguments. To that end, `a overlaps b` makes sense.

omus: I can change these back. Mainly I was thinking about these after changing `less_than_disjoint` to `isless_disjoint` and was thinking these should also start with `is` for consistency.

omus: Votes are to leave this as `overlaps`?

src/interval.jl
omus: Based upon [Alex's comment](https://github.com/invenia/Intervals.jl/pull/42/commits/89216aa06a69a344d8d8de68ff2811278eda6260#r203103948) should this be `contiguouswith`? For example `a contiguouswith b`.

omus: Alternatively: `adjacentto`

omus:
I'm going to move the "is" changes to another PR as they seem more controversial

src/interval.jl
ararslan: If we were to go with the "adjacent" terminology rather than "contiguous," I think `adjacent` itself is a fine enough name. That said, I don't really have a problem with `contiguous`, but I find `continugouswith` to be a bit of a mouthful.

PR #43, opened 2018-07-16
Compatibility with Julia 0.7-beta

omus:


omus:
Looks to be a coverage failure on 32-bit Windows for the Julia 0.7 nightly

omus:
We don't require coverage on Windows here but probably Coverage.jl should be tested on Windows.

A brief investigation reveals that the problem is from [this line](https://github.com/JuliaCI/Coverage.jl/blob/591ba4a78c3c7529b10b5470a9bfdf0a265bf911/src/parser.jl#L41):
```julia
Meta.parse(read(io, String), position(io)+1)
```
As it turns out `position` will always return a `Int64` and `parse(::AbstractString, ::Int)` will only take a `Int32` on 32-bit. We should ben able to fix this by modifying `parse` to take an `Integer` instead.

PR #44, opened 2018-07-17
Implement `eltype`

omus:
While writing the tests I noticed there wasn't a constructor which handled promotion to the `Interval` element type (e.g. `Interval{Float64}(1,2)`) so I also implemented that and in the process cleaned up the constructors.

### DIFF
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -14,6 +14,8 @@ import Base: ⊆, ⊇, ⊈, ⊉, union, union!, merge
 
 abstract type AbstractInterval{T} end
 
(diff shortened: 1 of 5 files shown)


omus:
32-bit issues on AppVeyor with Coverage. Intervals test all pass.

iamed2:
What is eltype of Interval actually used for?

omus:
I found myself wanting `eltype(::Interval)` in case where I was performing a `DataFrame` join using `Interval{ZonedDateTime}`s and I needed to pre-convert the time zones to be the same.

PR #45, opened 2018-07-17
Revisit `(less|greater)_than_disjoint` functions

omus:
Changes suggested by @ararslan (https://github.com/invenia/Intervals.jl/pull/42#discussion_r202523788).

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -170,24 +170,21 @@ end
 Base.isless(a::AbstractInterval, b) = LeftEndpoint(a) < b
 Base.isless(a, b::AbstractInterval) = a < LeftEndpoint(b)
 
-less_than_disjoint(a::AbstractInterval, b) = RightEndpoint(a) < b
-less_than_disjoint(a, b::AbstractInterval) = a < LeftEndpoint(b)
-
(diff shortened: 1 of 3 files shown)


src/interval.jl
```diff
@@ -170,24 +170,21 @@ end
...
-less_than_disjoint(a::AbstractInterval, b) = RightEndpoint(a) < b
-less_than_disjoint(a, b::AbstractInterval) = a < LeftEndpoint(b)
-
-function Base.:isless(a::AbstractInterval, b::AbstractInterval)
+function Base.isless(a::AbstractInterval, b::AbstractInterval)
     return LeftEndpoint(a) < LeftEndpoint(b)
 end
 
-function less_than_disjoint(a::AbstractInterval, b::AbstractInterval)
+isless_disjoint(a::AbstractInterval, b) = RightEndpoint(a) < b
```
iamed2: If it's called `isless`-something it should call `isless`, not `<`.

omus: Good point. I need to some investigation here.

omus:
I have some additional ideas regarding `isless` and `Interval`s I want to explore.

PR #47, opened 2018-07-25
Define hash methods

omus:
Addresses this problem:
```julia
julia> using Intervals, TimeZones

julia> n = now(tz"America/Winnipeg")
2018-07-25T15:06:54.901-05:00

julia> a = Interval(n, n)
Interval{TimeZones.ZonedDateTime}(2018-07-25T15:06:54.901-05:00, 2018-07-25T15:06:54.901-05:00, Inclusivity(true, true))

julia> b = deepcopy(a)
Interval{TimeZones.ZonedDateTime}(2018-07-25T15:06:54.901-05:00, 2018-07-25T15:06:54.901-05:00, Inclusivity(true, true))

julia> hash(a) == hash(b)
false

julia> isequal(a, b)
true
```

### DIFF
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -119,6 +119,15 @@ function Base.copy(x::AnchoredInterval{P, T}) where {P, T}
     return AnchoredInterval{P, T}(anchor(x), inclusivity(x))
 end
 
+function Base.hash(interval::AnchoredInterval{P}, h::UInt) where P
+    # Note: we shouldn't need to hash the second type parameter of an AnchoredInterval as
+    # that information should be covered by the anchor field.
+    h = hash(:AnchoredInterval, h)
+    h = hash(P, h)
+    h = hash(interval.anchor, h)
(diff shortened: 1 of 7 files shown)


rofinn:
Wouldn't it be easier to just use AutoHashEquals.jl?

omus:
> Wouldn't it be easier to just use AutoHashEquals.jl?

I usually use AutoHashEquals but in packages with very few dependencies I try to avoid it if it's simple enough.

omus:
I should have ran the full test suite locally. Stand by.

omus:
Ok, so there was a slight wrinkle with `AnchoredInterval` which has parameterized information which we need to include in the hash. Since AutoHashEquals doesn't deal with the parametric values I'm just doing our own thing.

omus:
The last commit changes how hashing works so that it is possible that `Interval` and `AnchoredInterval` types can hash to the same value. This reflects how Julia hashes number. e.g. `hash(1//2) == hash(0.5)`

PR #49, opened 2018-08-13
`isequal` fix

### LINKED ISSUE #48: Define `isequal`
```
Using Intervals

julia> NaN == NaN
false

julia> isequal(NaN, NaN)
true

julia> Interval(NaN, NaN) == Interval(NaN, NaN)
false  # This seems like the correct behaviour

julia> isequal(Interval(NaN, NaN), Interval(NaN, NaN))
false  # This should return true according to how `==` vs `isequal` work.
```

rofinn:
Closes #48

### DIFF
--- a/test/comparisons.jl
+++ b/test/comparisons.jl
@@ -1,4 +1,4 @@
-using Intervals: Ending, Beginning, overlaps, contiguous
+using Intervals: Ending, Beginning, overlaps, contiguous, RightEndpoint, LeftEndpoint
 
 function unique_paired_permutation(v::Vector{T}) where T
     results = Tuple{T, T}[]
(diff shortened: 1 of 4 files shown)


test/comparisons.jl
```diff
@@ -406,6 +406,15 @@ const INTERVAL_TYPES = [Interval, AnchoredInterval{Ending}, AnchoredInterval{Beg
         @test superset([a, b]) == Interval(1, 5, true, true)
     end
 
+    @testset "equal -0.0/0.0" begin
+        a = Interval(0.0, -0.0)
+        b = Interval(-0.0, 0.0)
```
omus: Can you do:
```julia
a = convert(A, Interval(...))
b = convert(B, Interval(...))
```

test/comparisons.jl
```diff
@@ -406,6 +406,15 @@ const INTERVAL_TYPES = [Interval, AnchoredInterval{Ending}, AnchoredInterval{Beg
...
+
+        @test a == b
+        @test !isequal(a, b)
+        @test hash(a) != hash(b)
```
omus: Should add the other comparisons to ensure they work as expected

rofinn: The endpoint comparisons?

omus: The other comparisons which are performed in this file like `<` and so on

PR #51, opened 2018-08-14
Fixes for non-ordered intervals

omus:


src/anchoredinterval.jl
```diff
@@ -121,8 +121,22 @@ end
 
 ##### ACCESSORS #####
 
-Base.first(interval::AnchoredInterval{P}) where P = min(interval.anchor, interval.anchor+P)
-Base.last(interval::AnchoredInterval{P}) where P = max(interval.anchor, interval.anchor+P)
+function Base.first(interval::AnchoredInterval{P}) where P
+    if P < zero(P)
```
rofinn: I'd write this as a ternary I guess.

omus: I thought it was visually too busy with the `+`:
```julia
function Base.first(interval::AnchoredInterval{P}) where P
    P < zero(P) ? interval.anchor + P : interval.anchor
end
```

ararslan: You could alternatively write it as `interval.anchor + P * (P < zero(P))` :stuck_out_tongue:

src/anchoredinterval.jl
```diff
@@ -223,6 +223,10 @@ end
 
 ##### EQUALITY #####
 
+function Base.:(==)(a::AnchoredInterval{P, T}, b::AnchoredInterval{P, T}) where {P, T}
```
rofinn: We should implement `isequal` here as well.

src/anchoredinterval.jl
rofinn: Also, we may want to just store the `extrema` as a tuple in the immutable type rather than performing this check every time.

omus: This is more complicated than it first seems see: https://github.com/invenia/Intervals.jl/issues/52

src/anchoredinterval.jl
omus: We can safely fall back to `isequal(::AbstractInterval, ::AbstractInterval)` here. I'll add some tests though to make sure we have the proper behaviour.

src/anchoredinterval.jl
omus: I'll go with a ternary statement with parameters around it. @ararslan's suggestion is good but there's an issue we need to solve with `LaxZonedDateTimes` which can result in the answer being wrong with you add a `TimePeriod` of 0...

ararslan: I was joking, really. Using a conditional, be it `if` or ternary, seems better to me.

src/anchoredinterval.jl
```diff
@@ -121,8 +121,14 @@ end
 
 ##### ACCESSORS #####
 
-Base.first(interval::AnchoredInterval{P}) where P = min(interval.anchor, interval.anchor+P)
-Base.last(interval::AnchoredInterval{P}) where P = max(interval.anchor, interval.anchor+P)
+function Base.first(interval::AnchoredInterval{P}) where P
```
rofinn: Don't forget to add a comment on why `min` and `max` won't work :)

rofinn:
Alright, I'm fine waiting to store the extrema in the type till later. LGTM.

src/anchoredinterval.jl
```diff
@@ -121,6 +121,12 @@ end
 
 ##### ACCESSORS #####
 
+# Note: For `first` and `last` the typically we would compute this is using `min` and `max`.
```
rofinn: "We would typically compute `first` and `last` using `min` and `max` respectively, but we can get unexpected behaviour if adding the span to the endpoint produces a value that is no longer comparable (e.g., `NaN`)."?

PR #53, opened 2018-10-17
Fix print errors on TimeTypes in 1.0

morris25:


src/description.jl
```diff
@@ -17,6 +17,10 @@ function description(interval::AnchoredInterval{P, ZonedDateTime}, s::String) wh
     )
 end
 
+function description(dt::TimeType, p::Period, suffix::String)
+    return "$(prefix(p))$suffix $(string(dt))"
```
ararslan: This might be a little more readable (IMO) as
```julia
string(prefix(p), suffix, ' ', dt)
```

src/description.jl
```diff
@@ -17,6 +17,10 @@ function description(interval::AnchoredInterval{P, ZonedDateTime}, s::String) wh
     )
 end
 
+function description(dt::TimeType, p::Period, suffix::String)
+    return string(prefix(p), suffix, " ", string(dt))
```
ararslan: You actually don't need the interior call to `string` here; each argument to `string` will be converted as though it itself was wrapped in `string`.

src/description.jl
```diff
@@ -32,6 +36,8 @@ function description(dt::AbstractDateTime, p::Period, suffix::String)
         ts = string(max_val)
     end
 
+    isa(dt, Time) && return "$(prefix(p))$suffix $ts"
```
ararslan: I'd similarly make this `string(prefix(p), suffix, ' ', ts)` for consistency, though this case is easier to read than the previous.

omus:
I'm not sure we should be supporting `Time` as an interval type. I think the motivation with using `AbstractDateTime` instead of `TimeType` was to not support `Time`. How did you notice this?

src/description.jl
morris25: whoops, missed that. Thanks!

morris25:
`LaxZonedDateTime` is a subtype of `TimeType` and one of our debugging statements was crashing because of it.
It might actually be better to fix the `print` method to use an `AbstractDateTime` instead of a `TimeType`, but that breaks a couple existing tests and we were using TimeType everywhere else in the code.

src/description.jl
```diff
@@ -32,6 +32,10 @@ function description(dt::AbstractDateTime, p::Period, suffix::String)
         ts = string(max_val)
     end
 
+    if VERSION < v"0.7.0-DEV.3216"
```
ararslan: Typically when inside of a function, `VERSION` checks should be wrapped in `@static`.

morris25: Thanks, I was unaware of that!

ararslan: Basically, when you use `@static`, the given condition is partially evaluated when the expression is parsed, and only the code in the conditional path taken is compiled when the function is compiled. If `@static` isn't used, the entire branch expression is compiled and the version check will be performed on every call (assuming the compiler doesn't optimize it out).

Issue #54, opened 2018-10-23
Generalize Endpoint


omus:
Something I noticed:
```julia
julia> zdt = ZonedDateTime(2018,10,23,14,tz"America/Winnipeg")
2018-10-23T14:00:00-05:00

julia> interval = HourEnding(zdt)
Intervals.AnchoredInterval{-1 hour,TimeZones.ZonedDateTime}(2018-10-23T14:00:00-05:00, Inclusivity(false, true))

julia> zdt <= Intervals.RightEndpoint(interval)
true

julia> zdt == Intervals.RightEndpoint(interval)
false

julia> zdt < Intervals.RightEndpoint(interval)
false
```
It is strange that `==` isn't working here. Additionally, these would also be nice:

- `Endpoint{Intervals.Left}(::AbstractInterval)` constructor
- `const First = Left; const Last = Right` as left and right may not be the most sensible and these match the `first` and `last` functions.

PR #56, opened 2018-10-25
Fix Scalar and Endpoint equality

fchorney:
This fixes the "==" issue in #54 
Also adds the Left/Right consts. 

- Added == functions for (left,left), (right,right), (scalar,left),
(scalar, right), (left, scalar), (right, scalar)
- Added isequal functions for (left,left) and (right,right)
- Changed some functions to not be repetitions of previous code such as
Left == Right code will just call Right == Left instead of redefining
the same code within both functions.
- Added First = Left, and Last = Right shortcuts.
- Added some more Scalar tests because why not

### DIFF
--- a/src/endpoint.jl
+++ b/src/endpoint.jl
@@ -6,9 +6,6 @@ const Right = Direction{:Right}()
 const Beginning = Left
 const Ending = Right
 
-const First = Left
-const Last = Right
-
 struct Endpoint{T, D}
     endpoint::T
     included::Bool
@@ -60,14 +57,6 @@ function Base.:(==)(a::Endpoint, b::Endpoint)
     a.endpoint == b.endpoint && a.included == b.included
 end
 
-function Base.:(==)(a::LeftEndpoint, b::LeftEndpoint)
-    a.endpoint == b.endpoint && a.included == b.included
-end
-
-function Base.:(==)(a::RightEndpoint, b::RightEndpoint)
-    a.endpoint == b.endpoint && a.included == b.included
-end
-
 function Base.:(==)(a::LeftEndpoint, b::RightEndpoint)
     a.endpoint == b.endpoint && a.included && b.included
 end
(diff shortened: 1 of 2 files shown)


src/endpoint.jl
```diff
@@ -58,24 +60,56 @@ function Base.:(==)(a::Endpoint, b::Endpoint)
...
 end
 
+function Base.:(==)(a::LeftEndpoint, b::LeftEndpoint)
+    a.endpoint == b.endpoint && a.included == b.included
+end
+
+function Base.:(==)(a::RightEndpoint, b::RightEndpoint)
+    a.endpoint == b.endpoint && a.included == b.included
+end
+
```
omus: Both of these are covered by falling back to `Base.:(==)(a::Endpoint, b::Endpoint)`

src/endpoint.jl
```diff
@@ -58,24 +60,56 @@ function Base.:(==)(a::Endpoint, b::Endpoint)
...
+    a == b.endpoint && b.included
+end
+
+function Base.:(==)(a::LeftEndpoint, b)
+    b == a
+end
+
+function Base.:(==)(a::RightEndpoint, b)
+    b == a
 end
```
omus: Can just use `Endpoint` here:
```julia
function Base.:(==)(a, b::Endpoint)
    a == b.endpoint && b.included
end

function Base.:(==)(a::Endpoint, b)
    b == a
end
```
I assume the one-liner versions of these look out of place?

src/endpoint.jl
```diff
@@ -58,24 +60,56 @@ function Base.:(==)(a::Endpoint, b::Endpoint)
...
     isequal(a.endpoint, b.endpoint) && isequal(a.included, b.included)
 end
 
+function Base.isequal(a::LeftEndpoint, b::LeftEndpoint)
+    isequal(a.endpoint, b.endpoint) && isequal(a.included, b.included)
+end
+
+function Base.isequal(a::RightEndpoint, b::RightEndpoint)
+    isequal(a.endpoint, b.endpoint) && isequal(a.included, b.included)
+end
```
omus: Fall back covers this

src/endpoint.jl
omus: These should probably be moved to the end with the `isless(::Any, ::Endpoint)` definitions. Since the logic behind scalar-endpoint and endpoint-endpoint comparisons is rather different I think it makes sense to group by this way instead of by operator.

src/endpoint.jl
```diff
@@ -6,6 +6,8 @@ const Right = Direction{:Right}()
 const Beginning = Left
 const Ending = Right
 
+const First = Left
+const Last = Right
```
omus: I would save this for another PR.

src/endpoint.jl
fchorney: Thats what I wanted to assume, but I was getting these weird issues where it seemed like it wasnt. Perhaps I was somewhat misunderstanding what I was doing. Tests definitely seem to run fine without them as they are falling back as you mentioned.

src/endpoint.jl
fchorney: Right, that makes sense. I suppose some of these could be one liners, but then the question is if we want it to be consistently long versions for the entire thing, or mixed, or make them all one liners. Some are little long though

Issue #60, opened 2019-01-11
failure with different types


sbromberger:
```
a = 0.0..0.2
3.0 in a # works
3 in a # fails with ERROR: MethodError: no method matching iterate(::Interval{Float64})
```

omus:
Fix included in Intervals 0.4.1

sbromberger:
Thank you!

Issue #59, opened 2019-01-04
Mirror Dates printing changes


omus:
Change introduced in: https://github.com/JuliaLang/julia/pull/30200

PR #61, opened 2019-01-15
Allow `in` with mixed types

omus:
Fixes: https://github.com/invenia/Intervals.jl/issues/60


omus:
Appveyor failure is: https://github.com/invenia/Intervals.jl/issues/59

PR #63, opened 2019-01-18
Make Inclusivity broadcastable as a scalar

morris25:


test/inclusivity.jl
```diff
@@ -47,6 +47,9 @@
         @test hash(inc) == hash(cp)
         @test !isequal(inc, diff)
         @test hash(inc) != hash(diff)
+
+        # Verify that Inclusivity is treated as a scalar during broadcast
+        Inclusivity(false, true) .== Inclusivity(false, true)
```
omus:
```suggestion
        @test size(Inclusivity(false, true) .== Inclusivity(false, true)) == ()
```

morris25: Thanks for catching that! I can't believe I didn't see it was missing.

test/inclusivity.jl
```diff
@@ -47,6 +47,9 @@
         @test hash(inc) == hash(cp)
         @test !isequal(inc, diff)
         @test hash(inc) != hash(diff)
+
+        # Verify that Inclusivity is treated as a scalar during broadcast
+        @test Inclusivity(false, true) .== Inclusivity(false, true)
```
omus: Something brought up by @iamed2 here: https://github.com/JuliaTime/TimeZones.jl/pull/177#issuecomment-456118302

```suggestion
        @test size(Inclusivity(false, true) .== Inclusivity(false, true)) == ()
```

PR #68, opened 2019-02-06
Update documentation to use doctests

omus:


.gitignore
```diff
@@ -4,3 +4,4 @@
 *.jl.mem
 /docs/build/
 /docs/site/
+Manifest.toml
```
ararslan: Perhaps add `docs/Manifest.toml` as well?

omus: This rule ignores all files named `Manifest.toml`. `/Manifest.toml` would be how to ignore only the root file.

ararslan: Oh whoa, TIL!

omus:
> Approved assuming the resulting docs look good

The docs could not look any better ;)

Issue #73, opened 2019-06-12
`Intervals.union` and `Intervals.union!` don’t give the same output


morris25:
Probably a problem specific to arrays of AnchoredIntervals
```
julia> target
10-element Array{AnchoredInterval{-1 hour,ZonedDateTime},1}:
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 14, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 14, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 8, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 8, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 8, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 8, tz"UTC"), Inclusivity(false, true))

julia> a = union(target)
3-element Array{AbstractInterval,1}:
 Interval{ZonedDateTime}(ZonedDateTime(2013, 2, 12, 6, tz"UTC"), ZonedDateTime(2013, 2, 12, 8, tz"UTC"), Inclusivity(false, true))
[...]

PR #74, opened 2019-06-14
Fix union and union! of intervals

iamed2:
For the purposes of these definitions, a "set" of intervals is either a single `AbstractInterval` or a vector of `AbstractInterval`s. 

This now supports the methods that set `union` supports. Notably, the method @morris25 found will now throw an `ArgumentError` rather than falling back to the Base definition and giving a different result.

Before:
```
julia> target
10-element Array{AnchoredInterval{-1 hour,ZonedDateTime},1}:
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 14, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 14, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 8, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 8, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 8, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 8, tz"UTC"), Inclusivity(false, true))

julia> a = union(target)
3-element Array{AbstractInterval,1}:
 Interval{ZonedDateTime}(ZonedDateTime(2013, 2, 12, 6, tz"UTC"), ZonedDateTime(2013, 2, 12, 8, tz"UTC"), Inclusivity(false, true))
 Interval{ZonedDateTime}(ZonedDateTime(2013, 2, 13, 6, tz"UTC"), ZonedDateTime(2013, 2, 13, 8, tz"UTC"), Inclusivity(false, true))
 Interval{ZonedDateTime}(ZonedDateTime(2013, 2, 14, 6, tz"UTC"), ZonedDateTime(2013, 2, 14, 7, tz"UTC"), Inclusivity(false, true))

julia> b = union!(target)
5-element Array{AnchoredInterval{-1 hour,ZonedDateTime},1}:
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 14, 7, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 12, 8, tz"UTC"), Inclusivity(false, true))
 AnchoredInterval{-1 hour,ZonedDateTime}(ZonedDateTime(2013, 2, 13, 8, tz"UTC"), Inclusivity(false, true))
```
After:
```
julia> a = union(target)
3-element Array{AbstractInterval,1}:
 Interval{ZonedDateTime}(ZonedDateTime(2013, 2, 12, 6, tz"UTC"), ZonedDateTime(2013, 2, 12, 8, tz"UTC"), Inclusivity(false, true))
 Interval{ZonedDateTime}(ZonedDateTime(2013, 2, 13, 6, tz"UTC"), ZonedDateTime(2013, 2, 13, 8, tz"UTC"), Inclusivity(false, true))
 Interval{ZonedDateTime}(ZonedDateTime(2013, 2, 14, 6, tz"UTC"), ZonedDateTime(2013, 2, 14, 7, tz"UTC"), Inclusivity(false, true))

julia> b = union!(target)
ERROR: ArgumentError: Cannot `union!` array of intervals of type AnchoredInterval{-1 hour,ZonedDateTime} as the type may change
Stacktrace:
 [1] union!(::Array{AnchoredInterval{-1 hour,ZonedDateTime},1}) at /Users/ericdavies/.julia/dev/Intervals/src/interval.jl:295
 [2] top-level scope at none:0
```

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -279,22 +279,24 @@ end
 
 # There is power in a union.
 """
-    union(intervals::AbstractVector{<:AbstractInterval})
+    union(intervals::Union{AbstractInterval, AbstractVector{<:AbstractInterval}}, itrs...)
 
-Flattens a vector of overlapping intervals into a new, smaller vector containing only
+Flattens sets of overlapping intervals into a new, smaller vector containing only
 non-overlapping intervals.
 """
-function Base.union(intervals::AbstractVector{<:AbstractInterval})
-    return union!(convert(Vector{AbstractInterval}, intervals))
+function Base.union(
+    interval::Union{AbstractInterval, AbstractVector{<:AbstractInterval}}, intervals...,
+)
+    return union!(AbstractInterval[interval; intervals...])
 end
 
-"""
-    union!(intervals::AbstractVector{<:Union{Interval, AbstractInterval}})
+function Base.union!(intervals::AbstractVector{T}) where T <: AbstractInterval
+    if !(T <: Interval) && T != AbstractInterval
+        throw(ArgumentError(
+            "Cannot `union!` array of intervals of type $T as the type may change"
+        ))
+    end
 
-Flattens a vector of overlapping intervals in-place to be a smaller vector containing only
-non-overlapping intervals.
-"""
-function Base.union!(intervals::Union{AbstractVector{<:Interval}, AbstractVector{AbstractInterval}})
     sort!(intervals)
 
     i = 2
@@ -317,6 +319,28 @@ function Base.union!(intervals::Union{AbstractVector{<:Interval}, AbstractVector
     end
 
     return intervals
+end
+
+"""
+    union!(intervals::AbstractVector{<:Union{Interval, AbstractInterval}}, itrs...)
+
+Flattens sets of overlapping intervals in-place into the first argument, making it a smaller
+vector containing only non-overlapping intervals.
+"""
+function union!(
+    intervals::AbstractVector{<:AbstractInterval},
+    interval_varargs...
+)
+    # imitates the behaviour of vcat/promote_eltypeof
+    for interval_vararg in interval_varargs
+        if interval_vararg isa AbstractInterval
+            push!(intervals, interval_vararg)
+        else
+            append!(intervals, interval_vararg)
+        end
+    end
+
+    return union!(intervals)
 end
 
 """
--- a/test/interval.jl
+++ b/test/interval.jl
@@ -536,6 +536,18 @@
     end
 
     @testset "union" begin
+        a = Interval(-100, -1)
+        b = Interval(-3, 10)
+
+        @test union(a) == AbstractInterval[a]
+        @test union(b) == AbstractInterval[b]
+        @test union(a, b) == AbstractInterval[Interval(-100, 10)]
+        @test union(b, a) == AbstractInterval[Interval(-100, 10)]
+        @test union([a, b]) == AbstractInterval[Interval(-100, 10)]
+        @test union([b, a]) == AbstractInterval[Interval(-100, 10)]
+        @test union!([a, b]) == [Interval(-100, 10)]
+        @test union!([b, a]) == [Interval(-100, 10)]
+
         intervals = [
             Interval(-100, -1, Inclusivity(false, false)),
             Interval(-10, -1, Inclusivity(false, false)),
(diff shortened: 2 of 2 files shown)


src/interval.jl
```diff
@@ -279,22 +279,24 @@ end
...
+Flattens sets of overlapping intervals into a new, smaller vector containing only
 non-overlapping intervals.
 """
-function Base.union(intervals::AbstractVector{<:AbstractInterval})
-    return union!(convert(Vector{AbstractInterval}, intervals))
+function Base.union(
+    interval::Union{AbstractInterval, AbstractVector{<:AbstractInterval}}, intervals...,
+)
+    return union!(AbstractInterval[interval; intervals...])
 end
```
omus: Seems odd we would support `union([Interval(1,2)], Interval(2,3))`

src/interval.jl
```diff
@@ -279,22 +279,24 @@ end
...
 
-"""
-    union!(intervals::AbstractVector{<:Union{Interval, AbstractInterval}})
+function Base.union!(intervals::AbstractVector{T}) where T <: AbstractInterval
+    if !(T <: Interval) && T != AbstractInterval
+        throw(ArgumentError(
+            "Cannot `union!` array of intervals of type $T as the type may change"
+        ))
+    end
```
omus: It would be better to use dispatch here for future interval types can support their own `union!` methods
```julia
function Base.union!(intervals::AbstractVector{<:AbstractInterval}) = 
    throw(ArgumentError("Cannot `union!` array of intervals of type $T as the type may change"))
end

Base.union!(intervals::AbstractVector{<:Union{AbstractInterval,Interval}) = ...
```
(I think this works)

omus:
Fixes: https://github.com/invenia/Intervals.jl/issues/73

src/interval.jl
iamed2: The problem with Intervals is that an Interval is just as much a set as a vector of intervals. It doesn't make sense to me to only allow vectors.

src/interval.jl
iamed2: Doesn't the existing definition already allow that?

src/interval.jl
omus: I was probably not very clear. I find it weird we are supporting mixing vectors of intervals and intervals. My preference would be to have:
```julia
Base.union(intervals::AbstractVector{<:AbstractInterval}) = ...
Base.union(intervals::AbstractInterval...) = ...
```

iamed2: My point still stands.

Also, vcat has the same behaviour.

omus: I'm not a fan of this behaviour but we should stick to how `union` currently behaves in `Base.

src/interval.jl
```diff
@@ -279,22 +279,24 @@ end
...
+    union(intervals::Union{AbstractInterval, AbstractVector{<:AbstractInterval}}, itrs...)
 
-Flattens a vector of overlapping intervals into a new, smaller vector containing only
+Flattens sets of overlapping intervals into a new, smaller vector containing only
 non-overlapping intervals.
 """
-function Base.union(intervals::AbstractVector{<:AbstractInterval})
-    return union!(convert(Vector{AbstractInterval}, intervals))
+function Base.union(
+    interval::Union{AbstractInterval, AbstractVector{<:AbstractInterval}}, intervals...,
```
omus: Should we do:
```suggestion
    interval::Union{AbstractInterval, AbstractVector{<:AbstractInterval}}, intervals::Union{AbstractInterval, AbstractVector{<:AbstractInterval}}...,
```
Avoids a call such as `union(Interval(1,2), 3)` causing a confusing error

src/interval.jl
omus: I think my point here was that this approach in the PR isn't the best for new interval types that want to define their own `union!` method.

PR #79, opened 2020-01-16
Drop 0.6

rofinn:
- Remove REQUIRE file
- Updating Project.toml with version number, description, license, etc.
- Remove Compat and releated deprecated references.


.appveyor.yml
```diff
@@ -1,41 +1,33 @@
...
-# matrix:
-#   allow_failures:
-#   - julia_version: nightly
-
+  - x86
+  - x64
+matrix:
+  allow_failures:
+    - julia_version: nightly
+    - platform: x86  # see https://github.com/invenia/JLSO.jl/issues/12
```
iamed2: Shouldn't we just delete this line since we're ignoring nightly failures on Travis altogether?

test/runtests.jl
```diff
@@ -1,28 +1,10 @@
...
-end
 
-# The name of the module is no longer printed for types reachable from Main
-if VERSION < v"0.7.0-DEV.2657"
-    mod_prefix = "Intervals."
-    tz_prefix = "TimeZones."
-else
-    mod_prefix = tz_prefix = ""
-end
+mod_prefix = tz_prefix = ""
```
iamed2: Can we just get rid of these instead?

.appveyor.yml
rofinn:
```suggestion
```

Project.toml
```diff
@@ -1,17 +1,18 @@
 name = "Intervals"
 uuid = "d8418881-c3e1-53bb-8760-2df7ec849ed5"
+license = "MIT"
+desc = "Non-iterable ranges"
```
mattBrzezinski: Where are these descriptions used? I can't seem to find anything on Google about them nor in the `Pkg.jl` [documentation](https://julialang.github.io/Pkg.jl/dev/toml-files/)

Project.toml
```diff
@@ -1,17 +1,18 @@
...
 authors = ["Curtis Vogt <curtis.vogt@gmail.com>", "Gem Newman <spurll@gmail.com>"]
```
nicoleepp: We should also update this to be "Invenia Technical Computing"

.travis.yml
```diff
@@ -4,30 +4,24 @@ os:
...
+jobs:
+  include:
+    - stage: Documentation
+      julia: 1.0
+      script: julia --project=docs -e '
+          using Pkg;
+          Pkg.develop(PackageSpec(path=pwd()));
+          Pkg.instantiate();
+          include("docs/make.jl");'
+      after_success: skip
```
omus: I'd just drop AppVeyor entirely. I understand if you don't want to do that as part of this PR.

Project.toml
omus: I'm still not sure how I feel about this. Listing the authors is helpful for figuring out who are the major contributors. I guess you could just use the contributors graph to get that info.

src/Intervals.jl
```diff
@@ -2,13 +2,11 @@ __precompile__()
...
-using Compat
-using Compat.Dates
-using Compat.Printf
+using Dates
+using Printf
 using TimeZones
 
-using Compat.Dates: value, coarserperiod
-using Compat: AbstractDateTime
+using Dates: value, coarserperiod, AbstractDateTime
```
omus: I think our preferred order has types first:
```suggestion
using Dates: AbstractDateTime, coarserperiod, value
```

test/anchoredinterval.jl
```diff
@@ -321,7 +321,7 @@ using Intervals: canonicalize
         ]
 
         for (interval, printed, shown) in tests
-            shown = mod_prefix * shown
+            shown = shown
```
omus:
```suggestion
```

Project.toml
rofinn: I think it's something that Pkg.jl developers talked about using to allow you to list packages in a registry including the description, but then never implemented. Didn't realize Pkg.jl just silently ignores it.

rofinn:
Are folks good if we merge and tag this?

omus:
> Are folks good if we merge and tag this?

One sec. We need to do some minor updates to Travis

.travis.yml
```diff
@@ -4,30 +4,24 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
   - 1.0
+  - 1.3
   - nightly
```
omus: Add 32-bit testing since we dropped AppVeyor
```suggestion
  - nightly
arch:
  - x64
  - x86
```

.travis.yml
```diff
@@ -4,30 +4,24 @@ os:
   - linux
   - osx
```
omus: Test on Windows too now
```suggestion
  - osx
  - windows
```

.travis.yml
```diff
@@ -4,30 +4,24 @@ os:
...
-#   - julia -e 'Pkg.clone(pwd()); Pkg.build("Intervals"); Pkg.test("Intervals"; coverage=true)'
 after_success:
-  - |
-    julia -e '
-      VERSION >= v"0.7.0-DEV.3656" && using Pkg
-      VERSION >= v"0.7.0-DEV.5183" || cd(Pkg.dir("Intervals"))
-      Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())
-      Pkg.add("Documenter"); include(joinpath("docs", "make.jl"))'
+  - julia -e 'using Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(process_folder())'
+jobs:
```
omus: Only do 32-bit on Linux:
```suggestion
jobs:
  # Test 32-bit only on Linux
  exclude:
    - arch: x86
      os: osx
    - arch: x86
      os: windows
      
```

.travis.yml
rofinn: Do we care about testing this on windows?

.travis.yml
rofinn: Is there anything we were concerned about being 64-bit specific?

omus: My thoughts were to keep this equivalent to how we had it setup before. After an offline discussion I think most of our repos can probably just test against 64-bit Linux and 32-bit Windows. In cases where there are failures we should support a more exhaustive CI matrix via a commit message trigger

PR #82, opened 2020-03-13
Add Plots recipes

oxinabox:
Take a look at the pictures added for the reference tests to see what it looks like.
Here is one of them:
This is for a collection of hour-ending anchored intervals. 
plot

The  x-axis is not intervals, but the type of the end points of the intervals.
Each interval is shown as a line segment between its end-points.

Not included in this PR:
 - Plotting just 1 interval, rather than a vector of them. (doesn't seem that useful)
 - Plotting intervals on the Y or Z axis (time is alway X axis to me)
 - Fancyness around letting people chose own end point markers

The end point markets of `:none` for exclusive and `|` for inclusive were chosen because they are compatible with all back-ends.

Got lots of help from @mkborregaard  with this.

PR #83, opened 2020-03-24
Go back to using AppVeyor to test windows

oxinabox:
Partially  reverts: https://github.com/invenia/Intervals.jl/pull/79

Need to do this because of:
https://github.com/JuliaIO/FFMPEG.jl/issues/14

This is so can test  #82

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,6 @@ environment:
   - julia_version: nightly
 
 platform:
-  - x86 # 32-bit
   - x64 # 64-bit
 
 matrix:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,41 @@
+environment:
+  matrix:
+  - julia_version: 1.0  # latest LTS
+  - julia_version: 1  # latest stable
+  - julia_version: nightly
+
+platform:
+  - x86 # 32-bit
+  - x64 # 64-bit
+
+matrix:
+  allow_failures:
+    - julia_version: nightly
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -0,0 +1,41 @@
+environment:
+  matrix:
+  - julia_version: 1.0  # latest LTS
+  - julia_version: 1  # latest stable
+  - julia_version: nightly
+
+platform:
+  - x86 # 32-bit
```
omus: Double testing on 32-bit again. I would just drop it from AppVeyor and do Linux 32-bit on Travis

omus:
AppVeyor needs to be re-enabled. Also, Travis didn't run?

appveyor.yml
oxinabox: Ah this time is is because I started from scratch in with AV
```suggestion
```

oxinabox:
> AppVeyor needs to be re-enabled. 

It doesn't look to me like it was ever disabled
https://ci.appveyor.com/project/invenia/intervals-jl

> Also, Travis didn't run?

weird, it's running now

oxinabox:
The trick used for TimeZones to not do plotting at all will not work for Intervals,
as we actually want to check what the plots look like in our tests,
not the attributes of them.

oxinabox:
What things should I do to complete this?
For some reasonm Travis is still showing pending, but I think it is done when I click on details.

PR #91, opened 2020-05-22
Overload timezone

glennmoy:
So we can extract the timezone directly from the `Interval`

```julia
TimeZones.timezone(x::Interval{ZonedDateTime}) = timezone(first(x))
TimeZones.timezone(x::AnchoredInterval{P, ZonedDateTime}) where P = timezone(anchor(x))
```

### DIFF
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -1,5 +1,3 @@
-import TimeZones: astimezone
-
 """
     AnchoredInterval{P, T}(anchor::T, [inclusivity::Inclusivity]) where {P, T} -> AnchoredInterval{P, T}
     AnchoredInterval{P, T}(anchor::T, [closed_left::Bool, closed_right::Bool]) where {P, T} -> AnchoredInterval{P, T}
@@ -284,3 +282,5 @@ canonicalize(target_type::Type{P}, p::P) where P <: Period = p
 function astimezone(i::AnchoredInterval{P, ZonedDateTime}, tz::TimeZone) where P
     return AnchoredInterval{P, ZonedDateTime}(astimezone(anchor(i), tz), inclusivity(i))
 end
+
+timezone(i::AnchoredInterval{P, ZonedDateTime}) where P = timezone(anchor(i))
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -279,8 +279,8 @@ canonicalize(target_type::Type{P}, p::P) where P <: Period = p
 
 ##### TIME ZONES #####
 
-function astimezone(i::AnchoredInterval{P, ZonedDateTime}, tz::TimeZone) where P
+function TimeZones.astimezone(i::AnchoredInterval{P, ZonedDateTime}, tz::TimeZone) where P
     return AnchoredInterval{P, ZonedDateTime}(astimezone(anchor(i), tz), inclusivity(i))
 end
 
-timezone(i::AnchoredInterval{P, ZonedDateTime}) where P = timezone(anchor(i))
+TimeZones.timezone(i::AnchoredInterval{P, ZonedDateTime}) where P = timezone(anchor(i))
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -348,6 +348,8 @@ function TimeZones.astimezone(i::Interval{ZonedDateTime}, tz::TimeZone)
 end
 
 function TimeZones.timezone(i::Interval{ZonedDateTime})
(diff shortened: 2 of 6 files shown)


nickrobinson251:
> we can extract the timezone directly from the Interval

What about when there are multiple timezones in the Interval? 

We can extract _a_ timezone, but not _the_ timezone. Arguably picking the timezone of the anchor in an AnchoredInterval is okay (although i'm not convinced), but picking either `timezone` in the below example feels arbitray to me. Is there a reason to favour `first`?
Personally, I think this is something the user should specify explicitly (using `first`,`last`, `anchor` etc).
```julia
julia> zdt1 = ZonedDateTime(2013, 2, 13, 0, 30, tz"America/Winnipeg");

julia> zdt2 = ZonedDateTime(2016, 8, 11, 21, tz"UTC");

julia> i = Interval(zdt1, zdt2)

julia> timezone(i)  # what should this return?
```

glennmoy:
> Is there a reason to favour first?

Hmm, I assumed you couldn't make an interval from two different timezones. Perhaps this is why it wasn't a feature already.

>this is something the user should specify explicitly

I think this is reasonable.

omus:
> What about when there are multiple timezones in the Interval?

I agree with @nickrobinson251. If there is a push to have this convenience it would be okay for an `Interval` to return *the* time zone if both endpoints use the same time zone otherwise throw an exception.

glennmoy:
> otherwise throw an exception.

@omus what type of exception would be appropriate here? I'm looking at `ErrorException` but it seems a bit generic?

omus:
> what type of exception would be appropriate here?

`ArgumentError` seems reasonable. The provided argument uses multiple time zones and is invalid for use in this function. If we need something even more specific we can make a custom exception but that only seems necessary if we come up with a use case.

src/anchoredinterval.jl
```diff
@@ -284,3 +282,5 @@ canonicalize(target_type::Type{P}, p::P) where P <: Period = p
 function astimezone(i::AnchoredInterval{P, ZonedDateTime}, tz::TimeZone) where P
     return AnchoredInterval{P, ZonedDateTime}(astimezone(anchor(i), tz), inclusivity(i))
 end
+
+timezone(i::AnchoredInterval{P, ZonedDateTime}) where P = timezone(anchor(i))
```
omus, line 286: We no longer like extending functions with `import` and now prefer:
```suggestion
TimeZones.timezone(i::AnchoredInterval{P, ZonedDateTime}) where P = timezone(anchor(i))
```

src/interval.jl
```diff
@@ -346,3 +346,5 @@ end
 function TimeZones.astimezone(i::Interval{ZonedDateTime}, tz::TimeZone)
     return Interval(astimezone(first(i), tz), astimezone(last(i), tz), inclusivity(i))
 end
+
+timezone(i::Interval{ZonedDateTime}) = timezone(first(i))
```
omus, line 350:
```suggestion
TimeZones.timezone(i::Interval{ZonedDateTime}) = timezone(first(i))
```

test/interval.jl
```diff
@@ -512,6 +512,13 @@
         end
     end
 
+    @testset "timezone" begin
+        zdt1 = ZonedDateTime(2013, 2, 13, 0, 30, tz"America/Winnipeg")
+        zdt2 = ZonedDateTime(2016, 8, 11, 21, tz"America/Winnipeg")
+        i = Interval(zdt1, zdt2)
```
omus, line 518: Add test for when time zones are different between endpoints

src/interval.jl
```diff
@@ -346,3 +346,8 @@ end
 function TimeZones.astimezone(i::Interval{ZonedDateTime}, tz::TimeZone)
     return Interval(astimezone(first(i), tz), astimezone(last(i), tz), inclusivity(i))
 end
+
+function TimeZones.timezone(i::Interval{ZonedDateTime})
+    timezone(first(i)) == timezone(last(i)) && return timezone(first(i))
+    throw(ArgumentError("Interval $i contains mixed timezones."))
```
omus, line 352: Definitely personal preference but I think this is a more standard way of writing this which I find easier to read:
```suggestion
    if timezone(first(i)) != timezone(last(i))
        throw(ArgumentError("Interval $i contains mixed timezones."))
    end
    return timezone(first(i))
```

glennmoy: yeah I guess throwing an error as the last line isn't very safe.

PR #92, opened 2020-05-25
Add unbounded endpoints. Represented by `nothing`

### LINKED ISSUE #89: Add support for unbounded intervals
https://www.postgresql.org/docs/9.3/rangetypes.html#RANGETYPES-INFINITE

fchorney:
Closes #89 

Initial WIP for adding unbounded endpoints which will be represented by `nothing`

There seems to be a lot of edge cases to take into account. There are still a lot of tests that need to be written, and more modifications to support unbounded intervals (such as intersection, union, etc.

### DIFF
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -41,5 +41,6 @@ export AbstractInterval,
        less_than_disjoint,
        greater_than_disjoint,
        superset,
-       .., ≪, ≫, ⊆, ⊇, ⊈, ⊉
+       .., ≪, ≫, ⊆, ⊇, ⊈, ⊉,
+       Endpoint, Left, Right, LeftEndpoint, RightEndpoint
 end
--- a/src/endpoint.jl
+++ b/src/endpoint.jl
@@ -103,36 +103,24 @@ function Base.isless(a::RightEndpoint, b::LeftEndpoint)
     return !isunbounded(a.endpoint) && !isunbounded(b.endpoint) && (a.endpoint < b.endpoint || (a.endpoint == b.endpoint && !(a.included && b.included)))
 end
 
+# Comparisons between Unbounded and Endpoints
+Base.isless(a::Endpoint{T}, b::Endpoint{T}) where T = a < b
+Base.isless(a::Endpoint{Nothing}, b::Endpoint{Nothing}) = false
+
+# Note: Are these two comparisons correct?
+Base.isless(a::Endpoint{Nothing}, b::Endpoint{T}) where T = true
+Base.isless(a::Endpoint{T}, b::Endpoint{Nothing}) where T = true
+
 # Comparisons between Scalars and Endpoints
 Base.:(==)(a, b::Endpoint) = a == b.endpoint && b.included
 Base.:(==)(a::Endpoint, b) = b == a
 
-Base.isless(a::Endpoint{T}, b::Endpoint{T}) where T = a < b
-Base.isless(a::Endpoint{Nothing}, b::Endpoint{T}) where T = true
-Base.isless(a::Endpoint{T}, b::Endpoint{Nothing}) where T = true
-Base.isless(a::Endpoint{Nothing}, b::Endpoint{Nothing}) = false
-
-function Base.isless(a, b::LeftEndpoint)
-    # If the left endpoint is unbounded, then there can be no value less than the endpoint
-    if isunbounded(b.endpoint)
-        return false
-    elseif isunbounded(a)
-        return true
-    else
-        return a < b.endpoint || (a == b.endpoint && !b.included)
-    end
-end
-
 Base.isless(a, b::RightEndpoint) = a < b.endpoint
 Base.isless(a::LeftEndpoint, b) = a.endpoint < b
+function Base.isless(a, b::LeftEndpoint)
+    return !isunbounded(b.endpoint) && (isunbounded(a) || (a < b.endpoint || (a == b.endpoint && !b.included)))
+end
+
 function Base.isless(a::RightEndpoint, b)
-    # If the right endpoint is unbounded, then there can be no value greater than the
-    # endpoint
-    if isunbounded(a.endpoint)
-        return false
-    elseif isunbounded(b)
-        return true
-    else
-        return a.endpoint < b || (a.endpoint == b && !a.included)
-    end
+    return !isunbounded(a.endpoint) && (isunbounded(b) || (a.endpoint < b || (a.endpoint == b && !a.included)))
 end
--- a/src/endpoint.jl
+++ b/src/endpoint.jl
@@ -7,10 +7,10 @@ const Beginning = Left
 const Ending = Right
 
 struct Endpoint{T, D}
-    endpoint::Union{T, Nothing}
+    endpoint::T
     included::Bool
 
-    function Endpoint{T, D}(ep::Union{T, Nothing}, included::Bool) where {T, D}
+    function Endpoint{T, D}(ep::T, included::Bool) where {T, D}
         @assert D isa Direction
         new{T, D}(ep, included)
     end
@@ -22,8 +22,14 @@ const RightEndpoint{T} = Endpoint{T, Right}
 LeftEndpoint(ep::T, included::Bool) where T = LeftEndpoint{T}(ep, included)
 RightEndpoint(ep::T, included::Bool) where T = RightEndpoint{T}(ep, included)
 
-LeftEndpoint(i::AbstractInterval{T}) where T = LeftEndpoint{T}(first(i), first(inclusivity(i)))
-RightEndpoint(i::AbstractInterval{T}) where T = RightEndpoint{T}(last(i), last(inclusivity(i)))
+function LeftEndpoint(i::AbstractInterval{T}) where T
+    X = isnothing(first(i)) ? Nothing : T
+    return LeftEndpoint{X}(first(i), first(inclusivity(i)))
+end
+function RightEndpoint(i::AbstractInterval{T}) where T
+    X = isnothing(last(i)) ? Nothing : T
+    return RightEndpoint{X}(last(i), last(inclusivity(i)))
+end
 
 function Base.hash(x::Endpoint{T, D}, h::UInt) where {T, D}
     # Note: we shouldn't need to hash `T` as this is covered by the endpoint field.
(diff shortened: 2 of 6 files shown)


omus:
> There are still a lot of tests that need to be written

I highly recommend you look at `test/comparisons.jl` for this. If you add all combinations of unbounded ranges to these tests you'll be in a really good state. You may not even need to write any tests beyond that

src/endpoint.jl
```diff
@@ -80,26 +80,59 @@ function Base.isequal(a::RightEndpoint, b::LeftEndpoint)
...
+function Base.isless(a, b::LeftEndpoint)
+    # If the left endpoint is unbounded, then there can be no value less than the endpoint
+    if isnothing(b.endpoint)
+        return false
+    else
+        return a < b.endpoint || (a == b.endpoint && !b.included)
+    end
+end
+# If the right endpoint is unbounded, then it is always greater than any other value
+Base.isless(a, b::RightEndpoint) = isnothing(b.endpoint) ? true : a < b.endpoint
```
omus, line 127:
```suggestion
Base.isless(a, b::RightEndpoint) = isnothing(b.endpoint) || a < b.endpoint
```

src/endpoint.jl
```diff
@@ -80,26 +80,59 @@ function Base.isequal(a::RightEndpoint, b::LeftEndpoint)
...
+# If the left endpoint is unbounded, then it is always less than any other value
+Base.isless(a::LeftEndpoint, b) = isnothing(a.endpoint) ? true : a.endpoint < b
```
omus, line 129:
```suggestion
Base.isless(a::LeftEndpoint, b) = isnothing(a.endpoint) || a.endpoint < b
```

src/interval.jl
```diff
@@ -76,7 +129,16 @@ Interval{T}(f, l, x::Bool, y::Bool) where T = Interval{T}(f, l, Inclusivity(x, y
...
-Interval(f, l, inc...) = Interval(promote(f, l)..., inc...)
+Interval(::Nothing, ::Nothing, inc...) = Interval{Nothing}(nothing, nothing, inc...)
+function Interval(f, l, inc...)
+    if isnothing(f) || isnothing(l)
+        # If only one endpoint is nothing, use the type of the other endpoint
+        T = isnothing(f) ? typeof(l) : typeof(f)
+        return Interval{T}(f, l, inc...)
+    else
+        return Interval(promote(f, l)..., inc...)
+    end
```
omus, line 140: Use more dispatch:
```julia
Interval(f::Nothing, l::T, inc...) where T = Interval{T}(f, l, inc...)
Interval(f::T, l::Nothing, inc...) where T = Interval{T}(f, l, inc...)
```

src/interval.jl
```diff
@@ -197,7 +265,14 @@ function Base.:isless(a::AbstractInterval, b::AbstractInterval)
...
 function less_than_disjoint(a::AbstractInterval, b::AbstractInterval)
-    return RightEndpoint(a) < LeftEndpoint(b)
+    re = RightEndpoint(a)
+    le = LeftEndpoint(b)
+
+    if isnothing(re) || isnothing(le)
+        return false
+    else
+        return re < le
+    end
```
omus, line 275: Logic shouldn't have to change here as endpoint comparisons should handle unbounded right?

src/interval.jl
```diff
@@ -241,7 +316,17 @@ true
...
+function Base.in(a, b::AbstractInterval)
+    if isnothing(a)
+        if isnothing(first(b)) && isnothing(last(b))
+            return true
+        else
+            return false
+        end
+    else
+        return !(a ≫ b || a ≪ b)
+    end
```
omus, line 328: I would expect there should be no changes here and this logic would be handled by `less_than_disjoint`

src/endpoint.jl
```diff
@@ -80,26 +80,59 @@ function Base.isequal(a::RightEndpoint, b::LeftEndpoint)
...
+    else
+        return a.endpoint < b.endpoint || (a.endpoint == b.endpoint && !(a.included && b.included))
+    end
 end
 
 # Comparisons between Scalars and Endpoints
-Base.:(==)(a, b::Endpoint) = a == b.endpoint && b.included
+function Base.:(==)(a, b::Endpoint)
+    a == b.endpoint && b.included
+end
```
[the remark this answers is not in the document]
fchorney: I think I was going to add something in here, but then found the actual place I had to change. Will revert

src/interval.jl
fchorney: I'll have to double check. At some point I found that some of the tests were failing here without modifications, but I'll go back to it and make sure

src/endpoint.jl
```diff
@@ -80,26 +80,59 @@ function Base.isequal(a::RightEndpoint, b::LeftEndpoint)
 end
 
 function Base.isless(a::LeftEndpoint, b::LeftEndpoint)
-    a.endpoint < b.endpoint || (a.endpoint == b.endpoint && a.included && !b.included)
+    if isnothing(a.endpoint) && isnothing(b.endpoint)
```
nickrobinson251, line 83: all of the `isnothing(x)`s should `x === nothing` for reasons i don't pretend to understand but which is referenced in https://docs.julialang.org/en/v1/manual/performance-tips/#Checking-for-equality-with-a-singleton-1 (xref https://github.com/JuliaLang/julia/issues/35585)

omus: That's true. We may want to have a `isunbounded` abstraction as we may switch our internal representation

fchorney: ah interesting. Will do. I had read somewhere to use `isnothing(x)` but that makes sense

omus: Actually, I believe `isnothing(x)` should be as performant as `x === nothing` it's `x == nothing` that can be slightly slower

fchorney: I like the abstraction idea. I'll throw that in as well

src/Intervals.jl
```diff
@@ -41,5 +41,6 @@ export AbstractInterval,
        less_than_disjoint,
        greater_than_disjoint,
        superset,
-       .., ≪, ≫, ⊆, ⊇, ⊈, ⊉
+       .., ≪, ≫, ⊆, ⊇, ⊈, ⊉,
+       Endpoint, Left, Right, LeftEndpoint, RightEndpoint
```
fchorney, line 45: I added this for testing purposes. I will need to remove this later.

src/endpoint.jl
```diff
@@ -80,26 +89,39 @@ function Base.isequal(a::RightEndpoint, b::LeftEndpoint)
...
+    return !isunbounded(a.endpoint) && !isunbounded(b.endpoint) && (a.endpoint < b.endpoint || (a.endpoint == b.endpoint && !(a.included && b.included)))
 end
 
+# Comparisons between Unbounded and Endpoints
+Base.isless(a::Endpoint{T}, b::Endpoint{T}) where T = a < b
+Base.isless(a::Endpoint{Nothing}, b::Endpoint{Nothing}) = false
+
+# Note: Are these two comparisons correct?
+Base.isless(a::Endpoint{Nothing}, b::Endpoint{T}) where T = true
+Base.isless(a::Endpoint{T}, b::Endpoint{Nothing}) where T = true
```
fchorney, line 113: I can't decide if this makes sense. Especially since we don't define a negative unbounded, and positive unbounded.

src/endpoint.jl
```diff
@@ -80,26 +89,39 @@ function Base.isequal(a::RightEndpoint, b::LeftEndpoint)
...
-Base.isless(a::LeftEndpoint, b)  = a.endpoint < b
-Base.isless(a::RightEndpoint, b) = a.endpoint < b || (a.endpoint == b && !a.included)
+Base.isless(a::LeftEndpoint, b) = a.endpoint < b
+function Base.isless(a, b::LeftEndpoint)
+    return !isunbounded(b.endpoint) && (isunbounded(a) || (a < b.endpoint || (a == b.endpoint && !b.included)))
+end
+
+function Base.isless(a::RightEndpoint, b)
+    return !isunbounded(a.endpoint) && (isunbounded(b) || (a.endpoint < b || (a.endpoint == b && !a.included)))
+end
```
fchorney, line 127: Similar comment here as the one right above.

test/runtests.jl
```diff
@@ -8,10 +8,10 @@ using VisualRegressionTests
 
 
 @testset "Intervals" begin
-    include("inclusivity.jl")
-    include("endpoint.jl")
+    #include("inclusivity.jl")
```
fchorney, line 11: I need to remember to add these back in

src/endpoint.jl
```diff
@@ -1,3 +1,6 @@
+# isnothing doesn't exist in Julia 1.0
+isunbounded(a) = a === nothing
```
nickrobinson251, line 2: calling this `isunbounded` rather than `isbounded` means the code has lots of double-negatives `!isunbounded` ...i wonder if flipping the condition would make the long chains of `&&` and `||` a little easier to read
```suggestion
isbounded(a) = a !== nothing
```

nickrobinson251:
This PR does a lot of work essentially use `nothing` as `Inf`. But we already have `Inf`.
```julia
julia> Interval(-Inf, 100.0)
Interval{Float64}(-Inf, 100.0, Inclusivity(true, true))

julia> Interval(100.0, Inf)
Interval{Float64}(100.0, Inf, Inclusivity(true, true))
```
Of course, `Inf isa Float64`, so for `Real`s it will promote the `Interval`s to be `Interval{Float64}`
```julia
julia> Interval(π, Inf)
Interval{Float64}(3.141592653589793, Inf, Inclusivity(true, true))

julia> Interval(10//2, Inf)
Interval{Float64}(5.0, Inf, Inclusivity(true, true))

julia> Interval(10, Inf)
Interval{Float64}(5.0, Inf, Inclusivity(true, true))
```
And for time types, there is no promotion
```julia
julia> using Dates

julia> Interval(now(), Inf)
ERROR: promotion of types DateTime and Float64 failed to change any arguments
```
But it does make me wonder if what we want here is just an "InfDateTime"?

omus:
> But it does make me wonder if what we want here is just an "InfDateTime"?

We are also looking into such an approach. As intervals only require the provided bounds to be orderable we'd need a representation of infinity for each type wanting to support unbounded endpoints. We're also looking into https://github.com/cjdoris/Infinity.jl but we could take other approaches like you suggest.

I will note however we do not want to promote to `Float64` for unbounded integer intervals:
```julia
julia> Interval(1, Inf)
Interval{Float64}(1.0, Inf, Inclusivity(true, true))
```
The Infinity.jl package approach is more suitable:
```julia
julia> Interval(1, ∞)
Interval{InfExtended{Int64}}(1, ∞, Inclusivity(true, true))
```

fchorney:
Yes, I'm thinking maybe at this point in this PR, I should maybe change my efforts to adding a Dates PR to Infinity.jl and just end up using that. There will still be a few changes needed for this package, but I think that would be much easier in the end. This is also assuming that the maintainer of Infinity.jl will be open to a Dates addition.

omus:
Replaced by #104. Note that using `Union{T,Nothing}` in the `struct` proposed here results in the type `Interval` no longer being a bits-type. This was worked around in #104 by always using `T` and using the type parameters to specify if the lower/upper bound was unbounded.

PR #95, opened 2020-05-27
Add support for Infinity.jl

fchorney:
Using #92 as a jump off point. This is a PR that utilizing Infinity.jl instead of using `nothing` as a sentinel value to represent unbounded ranges. 

There were still a few changes and fixes to be made, but I think this somewhat just works(tm). 

If this is good, the next steps would be to add Date/Time support to Infinity.jl

Most of this PR is just adding tests

### DIFF
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -1,8 +1,7 @@
 module Intervals
 
 using Dates
-using Infinity
-using Infinity: isposinf, isneginf
+using Infinity: Infinite, InfExtended, isposinf, isneginf
 using Printf
 using RecipesBase
 using TimeZones
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -2,6 +2,7 @@ module Intervals
 
 using Dates
 using Infinity
+using Infinity: isposinf, isneginf
 using Printf
 using RecipesBase
 using TimeZones
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -1,7 +1,6 @@
 module Intervals
 
 using Dates
-using Infinity: Infinite, InfExtended, isposinf, isneginf
 using Printf
 using RecipesBase
 using TimeZones
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -1,6 +1,7 @@
 module Intervals
 
 using Dates
+using Infinity
 using Printf
 using RecipesBase
 using TimeZones
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -1,6 +1,7 @@
 module Intervals
 
 using Dates
+using Infinity: InfExtended
 using Printf
 using RecipesBase
 using TimeZones
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -71,27 +71,21 @@ struct Interval{T} <: AbstractInterval{T}
     end
 end
 
-isbounded(a) = !isposinf(a) && !isneginf(a)
-isunbounded(a) = !isbounded(a)
-
-function Interval{T}(f, l, inc::Inclusivity) where T
-    if (isbounded(f) && isbounded(l)) || (isunbounded(f) && isunbounded(l))
-        return Interval{T}(convert(T, f), convert(T, l), inc)
+"""
+Check if a value is ∞ or not. Specifically check if the type is `Infinite`,
+or `InfExtended`, and then check if the value is pos or neg infinity.
+"""
+function isbounded(a)
+    T = typeof(a)
+    if T <: Infinite || T <: InfExtended
+        return !isposinf(a) && !isneginf(a)
     else
-        # If either endpoint is unbounded, we want to convert the bounded variable, and then
-        # try and promote them both to a compatable type.
-        # If T is a subset of the Infinite type, then don't try to convert at all, as trying
-        # to convert any type to Infinite will result in an error
-        if !(T <: Infinite)
-            if isbounded(f)
-                f = convert(T, f)
-            else
-                l = convert(T, l)
-            end
-        end
-        return Interval(f, l, inc)
+        return true
     end
 end
+isunbounded(a) = !isbounded(a)
+
+Interval{T}(f, l, inc::Inclusivity) where T = Interval{T}(convert(T, f), convert(T, l), inc)
 Interval{T}(f, l, x::Bool, y::Bool) where T = Interval{T}(f, l, Inclusivity(x, y))
 Interval{T}(f, l) where T = Interval{T}(f, l, true, true)
 
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -103,6 +103,14 @@ function Interval{T}(left::LeftEndpoint{T}, right::RightEndpoint{T}) where T
     Interval{T}(left.endpoint, right.endpoint, left.included, right.included)
 end
 
+"""
+When the endpoints are different types, try to find a common type to combined them into
+"""
+function Interval(left::LeftEndpoint{S}, right::RightEndpoint{D}) where {S, D}
+    T = promote_type(S, D)
+    return Interval{T}(left.endpoint, right.endpoint, left.included, right.included)
+end
+
 Interval(left::LeftEndpoint{T}, right::RightEndpoint{T}) where T = Interval{T}(left, right)
 
 # Empty Intervals
@@ -353,14 +361,7 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
 
     left = min(LeftEndpoint(a), LeftEndpoint(b))
     right = max(RightEndpoint(a), RightEndpoint(b))
-
-    # This promotion fixes the situation where one endpoint has a type of
-    # `InfExtended{T}` yet the ∞ value is of type `Infinite`.
-    # This will cause an error when trying to `promote(left, right)`
-    return Interval(
-        promote(left.endpoint, right.endpoint)...,
-        left.included, right.included
-    )
+    return Interval(left, right)
 end
 
 ##### TIME ZONES #####
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -71,18 +71,7 @@ struct Interval{T} <: AbstractInterval{T}
     end
 end
 
-"""
-Check if a value is ∞ or not. Specifically check if the type is `Infinite`,
-or `InfExtended`, and then check if the value is pos or neg infinity.
-"""
-function isbounded(a)
-    T = typeof(a)
-    if T <: Infinite || T <: InfExtended
-        return !isposinf(a) && !isneginf(a)
-    else
-        return true
-    end
-end
+isbounded(x) = isfinite(x)
 isunbounded(a) = !isbounded(a)
 
 Interval{T}(f, l, inc::Inclusivity) where T = Interval{T}(convert(T, f), convert(T, l), inc)
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -71,7 +71,27 @@ struct Interval{T} <: AbstractInterval{T}
     end
 end
 
-Interval{T}(f, l, inc::Inclusivity) where T = Interval{T}(convert(T, f), convert(T, l), inc)
+isbounded(a) = !isposinf(a) && !isneginf(a)
+isunbounded(a) = !isbounded(a)
+
+function Interval{T}(f, l, inc::Inclusivity) where T
+    if (isbounded(f) && isbounded(l)) || (isunbounded(f) && isunbounded(l))
+        return Interval{T}(convert(T, f), convert(T, l), inc)
+    else
+        # If either endpoint is unbounded, we want to convert the bounded variable, and then
+        # try and promote them both to a compatable type.
+        # If T is a subset of the Infinite type, then don't try to convert at all, as trying
+        # to convert any type to Infinite will result in an error
+        if !(T <: Infinite)
+            if isbounded(f)
+                f = convert(T, f)
+            else
+                l = convert(T, l)
+            end
+        end
+        return Interval(f, l, inc)
+    end
+end
 Interval{T}(f, l, x::Bool, y::Bool) where T = Interval{T}(f, l, Inclusivity(x, y))
 Interval{T}(f, l) where T = Interval{T}(f, l, true, true)
 
@@ -161,7 +181,8 @@ end
 
 ##### ARITHMETIC #####
 
-Base.:+(a::T, b) where {T <: Interval} = T(first(a) + b, last(a) + b, inclusivity(a))
+Base.:+(a::Interval, b) = Interval(first(a) + b, last(a) + b, inclusivity(a))
+
 
 Base.:+(a, b::Interval) = b + a
 Base.:-(a::Interval, b) = a + -b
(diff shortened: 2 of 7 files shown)


test/comparisons.jl
```diff
@@ -18,31 +18,39 @@ const INTERVAL_TYPES = [Interval, AnchoredInterval{Ending}, AnchoredInterval{Beg
...
-        @test superset([earlier, later]) == Interval(1, 5, true, true)
+        tests = [
+            ((1, 2, true, true), (4, 5, true, true)),
+            ((-∞, 2, true, true), (4, ∞, true, true)),
+        ]
+        for (a, b) in tests
+            @show a
+            @show b
+            earlier = convert(A, Interval(a[1], a[2], a[3], a[4]))
+            later = convert(B, Interval(b[1], b[2], b[3], b[4]))
```
fchorney, line 29: These converts seem to fail for AnchoredIntervals, and I currently can't seem to figure out how to fix it. 

Seems to fail here https://github.com/invenia/Intervals.jl/blob/master/src/anchoredinterval.jl#L165 with the error 

```
a = (-∞, 2, true, true)
b = (4, ∞, true, true)
non-overlapping: Error During Test at /Users/fchorney/Documents/repos/julia/Intervals.jl/test/comparisons.jl:20
  Got exception outside of a @test
  TypeError: in Type, in parameter, expected Int64, got InfExtended{Int64}
  Stacktrace:
   [1] convert(::Type{AnchoredInterval{Intervals.Direction{:Right}(),T} where T}, ::Interval{InfExtended{Int64}}) at /Users/fchorney/Documents/repos/julia/Intervals.jl/src/anchoredinterval.jl:167
   [2] top-level scope at /Users/fchorney/Documents/repos/julia/Intervals.jl/test/comparisons.jl:29
   [3] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
   [4] top-level scope at /Users/fchorney/Documents/repos/julia/Intervals.jl/test/comparisons.jl:21
   [5] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1180
   [6] include at ./boot.jl:328 [inlined]
   [7] include_relative(::Module, ::String) at ./loading.jl:1105
   [8] include(::Module, ::String) at ./Base.jl:31
   [9] include(::String) at ./client.jl:424
   [10] top-level scope at /Users/fchorney/Documents/repos/julia/Intervals.jl/test/runtests.jl:15
   [11] top-level scope at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.3/Test/src/Test.jl:1107
   [12] top-level scope at /Users/fchorney/Documents/repos/julia/Intervals.jl/test/runtests.jl:15
   [13] include at ./boot.jl:328 [inlined]
   [14] include_relative(::Module, ::String) at ./loading.jl:1105
   [15] include(::Module, ::String) at ./Base.jl:31
   [16] include(::String) at ./client.jl:424
   [17] top-level scope at none:6
   [18] eval(::Module, ::Any) at ./boot.jl:330
   [19] exec_options(::Base.JLOptions) at ./client.jl:263
   [20] _start() at ./client.jl:460
```

src/interval.jl
```diff
@@ -71,7 +71,27 @@ struct Interval{T} <: AbstractInterval{T}
...
+        if !(T <: Infinite)
+            if isbounded(f)
+                f = convert(T, f)
+            else
+                l = convert(T, l)
+            end
+        end
+        return Interval(f, l, inc)
+    end
+end
```
omus, line 94: I think if a user has explicitly set `T` we shouldn't be messing with it.

src/interval.jl
```diff
@@ -161,7 +181,8 @@ end
 
 ##### ARITHMETIC #####
 
-Base.:+(a::T, b) where {T <: Interval} = T(first(a) + b, last(a) + b, inclusivity(a))
+Base.:+(a::Interval, b) = Interval(first(a) + b, last(a) + b, inclusivity(a))
```
omus, line 184: The previous version always kept the `Interval` unit type the same. This could cause the Interval type to change.

src/interval.jl
```diff
@@ -338,7 +359,14 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
...
     right = max(RightEndpoint(a), RightEndpoint(b))
-    return Interval(left, right)
+
+    # This promotion fixes the situation where one endpoint has a type of
+    # `InfExtended{T}` yet the ∞ value is of type `Infinite`.
+    # This will cause an error when trying to `promote(left, right)`
+    return Interval(
+        promote(left.endpoint, right.endpoint)...,
+        left.included, right.included
+    )
```
omus, line 369: Seems preferable to change the endpoint constructor (`Interval{T}(left::LeftEndpoint{T}, right::RightEndpoint{T})`)

test/comparisons.jl
```diff
@@ -18,31 +18,41 @@ const INTERVAL_TYPES = [Interval, AnchoredInterval{Ending}, AnchoredInterval{Beg
...
-        @test isempty(intersect(earlier, later))
-        @test_throws ArgumentError merge(earlier, later)
-        @test union([earlier, later]) == [earlier, later]
-        @test !overlaps(earlier, later)
-        @test !contiguous(earlier, later)
-        @test superset([earlier, later]) == Interval(1, 5, true, true)
+        tests = [
+            ((1, 2, true, true), (4, 5, true, true)),
+            ((-∞, 2, true, true), (4, ∞, true, true)),
+        ]
```
omus, line 24: Testsets have a for-loop syntax which would be better as you can output details in the name of the testset (https://docs.julialang.org/en/v1/stdlib/Test/#Test.@testset). I'd probably prefer a distinct testset named "non-overlapping unbounded"

test/comparisons.jl
omus: I think this call is occuring `convert(AnchoredInterval{Ending}, Interval(4, ∞, true, true))` which doesn't make sense.

src/interval.jl
```diff
@@ -71,7 +71,27 @@ struct Interval{T} <: AbstractInterval{T}
     end
 end
 
-Interval{T}(f, l, inc::Inclusivity) where T = Interval{T}(convert(T, f), convert(T, l), inc)
+isbounded(a) = !isposinf(a) && !isneginf(a)
+isunbounded(a) = !isbounded(a)
```
omus, line 75: Should only apply to `AbstractInterval`s

src/Intervals.jl
```diff
@@ -1,6 +1,8 @@
 module Intervals
 
 using Dates
+using Infinity
```
omus, line 4: Is this needed as you explicitly import methods below?

test/interval.jl
```diff
@@ -180,6 +206,36 @@
...
+        @test !(Interval(-∞, ∞) > Interval(-∞, ∞))
+        @test Interval(100, 200) > Interval(-∞, ∞)
+        @test Interval(100, 200) > Interval(-∞, 150)
+        @test Interval(150, 200) > Interval(100, ∞)
+
+        # Equality
+        @test Interval(-∞, ∞) == Interval(-∞, ∞)
+        @test Interval(-∞, ∞) != Interval(-∞, ∞, true, false)
+        @test Interval(-∞, ∞) != Interval(-∞, ∞, false, true)
+        @test Interval(-∞, ∞) != Interval(-∞, ∞, false, false)
```
omus, line 236: I thought we weren't allowing inclusive endpoints for unbounded endpoints?

test/comparisons.jl
fchorney: I’m unsure why that doesn’t make any sense? Seems like it would work how it used to but now has infinity values in it

src/Intervals.jl
fchorney: If you don’t then you can’t use the infinity symbol correctly. I suppose I could just import it with the other items.

test/interval.jl
fchorney: I wasn’t sure if we still wanted to do that with proper infinity. As it potentially makes sense for infinity itself to be inclusive or not? It definitely makes the logic nicer to include inclusivity. I am willing to go back to mutating the inclusivity.

test/comparisons.jl
omus: The reason it's non-sense is the point at which you're setting the one endpoint used in the `AnchoredInterval` is `∞`. Then we're calculating the other endpoint by subtracting `4` from `∞` which is also `∞`. There doesn't seem to be a way to do this conversion without losing information which is why I'm thinking this conversion should error. We should detect this and report a nice error.

src/Intervals.jl
omus: In the context of the package I don't think we're referencing the infinity symbol

test/interval.jl
omus: As infinity is the first non-finite value we're supporting in intervals I think there is going to be some specialized logic around it. I think I'll write some of the logic down in the original issue.

src/Intervals.jl
fchorney: hmm good point I guess it's just for the tests.

test/interval.jl
omus: After writing down specific rules I think we can leave this as it is: https://github.com/invenia/Intervals.jl/issues/89#issuecomment-635372868

test/comparisons.jl
fchorney: ah ok I see. Thanks for the explanation!

src/interval.jl
fchorney: Hmm this is actually checking values rather than intervals, as it's mostly just to check if a value is set to ∞ or not. Maybe the name needs to change?

src/interval.jl
fchorney: hm yeah I suppose this change is actually just a byproduct of trying to throw in unbounded values into the tests.

This test is specifically the issue (https://github.com/invenia/Intervals.jl/blob/fc/using-infinity/test/interval.jl#L46-L47):
```
@test Interval(b, a, Inclusivity(true, false)) == Interval{typeof(a)}(a, b, Inclusivity(false, true))
```
When you create the interval on the left size of `==` it will `promote(b, a)` so if one value is `∞` and the other is a value, lets say `10`, then it converts both values to `InfExtended`. Now on the right side of `==`, if a is of type `Infinite` (due to being a `∞` or `-∞`) then it will try to `convert(Infinite, 10)` which is an error. On the other side of the spectrum, if a is `10` which would make it `Int64` type, then trying to do `convert(Int64, ∞)` will also fail. I guess we can just let it be an error if somebody puts in the wrong type in the Interval constructor.

omus: I think the test should be updated to be using:
```julia
julia> promote_type(Infinite, Int)
InfExtended{Int64}
```
Then in the tests you can use `Interval{promote_type(typeof(a), typeof(b))}(...)`

src/interval.jl
omus: My bad then. I think you can leave this as is for now.

src/interval.jl
fchorney: The problem here is that you'll end up with situations such that 
```
typeof(left) = Intervals.Endpoint{InfExtended{Int64},Intervals.Direction{:Left}()}
typeof(right) = Intervals.Endpoint{Int64,Intervals.Direction{:Right}()}
```
Thus you don't actually make it to that constructor you mention since the types are different. It ends up trying to call `Interval(f, l, inc...) = Interval(promote(f, l)..., inc...)` which fails because there are no rules on how to promote two endpoints.

fchorney: Would it make sense to also define a constructor for two endpoints where the types are different? I seems like we typically don't want to change what the Interval Type is, but with Infinity it seems to be required in certain situations

```
function Interval(left::LeftEndpoint{S}, right::RightEndpoint{D}) where {S, D}
    T = promote_type(S, D)
    return Interval{T}(left.endpoint, right.endpoint, left.included, right.included)
end
```

src/interval.jl
fchorney: This becomes the same issue as most of the other issues that cropped up by adding Infinity. If `a` is `-∞ .. ∞` and `b` is of type `Int` then `∞ - b` where `b` is `1` becomes an `InfExtended{Int64}` type which will error out since T would have been `Interval{Infinite}(...)`. So we end up trying to run `convert(Infinite, InfExtended{Int64})` which is undefined. That being said, maybe that shouldn't be undefined. Converting an `InfExtended{Int64})` to `Infinite` could potentially work if we check the value to make sure it's not an actual extended value

src/interval.jl
```diff
@@ -71,6 +71,20 @@ struct Interval{T} <: AbstractInterval{T}
...
+or `InfExtended`, and then check if the value is pos or neg infinity.
+"""
+function isbounded(a)
+    T = typeof(a)
+    if T <: Infinite || T <: InfExtended
+        return !isposinf(a) && !isneginf(a)
+    else
+        return true
+    end
+end
```
omus, line 85: `isfinite` works with `Infinity` and `InfExtended` types as well as the `Inf` from `Float64`. Additionally this works with `TimeType`s:
```suggestion
isbounded(x) = isfinite(x)
```

omus:
We were aware that `AnchoredInterval`s could only support representing an unbounded interval where the finite point would be stored. Since an `AnchoredInterval` uses a single point and a span to represent a range it is only possibly to represent a range in this form if we use finite point (due to `infinity + finite = infinity`).

However there is currently another problem:
```julia
julia> convert(AnchoredInterval{Beginning}, Interval(4, ∞, true, true))
ERROR: TypeError: in Type, in parameter, expected Int64, got InfExtended{Int64}
Stacktrace:
 [1] convert(::Type{AnchoredInterval{Intervals.Direction{:Left}(),T} where T}, ::Interval{InfExtended{Int64}}) at /Users/omus/.julia/dev/Intervals/src/anchoredinterval.jl:169
 [2] top-level scope at REPL[4]:1
```
This problem is due to `isbits(InfExtended{Int64}(∞))` being `false`:
```julia
julia> AnchoredInterval{InfExtended{Int64}(∞),InfExtended{Int64}}(4,Inclusivity(true,true))
ERROR: TypeError: in Type, in parameter, expected Int64, got InfExtended{Int64}
Stacktrace:
 [1] top-level scope at REPL[5]:1

julia> AnchoredInterval{∞,InfExtended{Int64}}(4,Inclusivity(true,true))
AnchoredInterval{∞,InfExtended{Int64}}(4, Inclusivity(true, true))
```

omus:
Solutions:
- Make `InfExtended{T}` a bitstype which allows values of this type to be used for parametric structs (https://github.com/cjdoris/Infinity.jl/issues/4)
- Eliminate using span values in the type parameter (https://github.com/invenia/Intervals.jl/pull/12)

Workarounds:
- Ignore/remove unbounded support for `AnchoredInterval`s

src/interval.jl
omus: That is one option. We may alternatively want to define our own `promote_rule`

src/interval.jl
omus: That conversion makes sense if the `InfExtended{Int64}` is infinite

omus: Fixed by adding the conversion: https://github.com/cjdoris/Infinity.jl/pull/7

omus:
Closing this PR as unbounded support was added in https://github.com/invenia/Intervals.jl/pull/104 and support for Infinity.jl was added in https://github.com/invenia/Intervals.jl/pull/107

Issue #57, opened 2018-11-28
Fail instead of re-ordering interval endpoints


omus:
Intervals enforces that the first endpoint always precedes the last endpoint:
```julia
julia> Interval(1, 2)
Interval{Int64}(1, 2, Inclusivity(true, true))

julia> Interval(2, 1)
Interval{Int64}(1, 2, Inclusivity(true, true))
```
This behaviour could be confusing when supplying inclusiveness:
```julia
julia> x, y = 2, 1;

julia> Interval(x, y, false, true)  # Should `false` associated with `x` or the first endpoint?
Interval{Int64}(1, 2, Inclusivity(true, false))
```
I would suggest we should deprecate this behaviour as it's already a corner case.

omus:
I'll point out that IntervalSets.jl don't enforce this and they have strange outcomes when this occurs:
```julia
julia> using IntervalSets

julia> Interval(1, 3) ∩ Interval(2, 3)
2..3

julia> Interval(3, 1) ∩ Interval(2, 3)
3..1
```

PR #97, opened 2020-06-02
Deprecate reordering of Interval endpoints

omus:
Adds deprecation for: https://github.com/invenia/Intervals.jl/issues/57

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -60,6 +60,14 @@ struct Interval{T, L <: Bound, R <: Bound} <: AbstractInterval{T,L,R}
         f, l, left_bound, right_bound = if f ≤ l
             f, l, L, R
         elseif l ≤ f
+            # Note: Including the full stacktrace in this deprecation warning as most calls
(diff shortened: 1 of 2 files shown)


src/interval.jl
```diff
@@ -62,6 +62,12 @@ struct Interval{T} <: AbstractInterval{T}
         f, l, inc = if f ≤ l
             f, l, inc
         elseif l ≤ f
+            Base.depwarn(
+                "Constructing an `Interval(x, y, ...)` is where `x > y` is deprecated, " *
+                "use `Interval(y, x, ...)` instead." *
+                sprint(Base.show_backtrace, stacktrace()),
```
omus, line 68: Normally `depwarn` just shows the first caller but this was inadequate as most calls to this constructor are from an `Interval` constructor.

nicoleepp:
Has conflicts now

omus:
I was going to get this in after #104. I can update this PR in advance of that.

PR #99, opened 2020-06-03
Support intervals endpoints using infinity

omus:
Adapted from https://github.com/invenia/Intervals.jl/pull/95. Takes the, hopefully, non-controversial parts of that PR and uses the `Float64` infinity to validate the changes.

@fchorney doing this allowed me to go very carefully through your tests. I've removed a few of them that are redundant with those which are in comparison.jl and some were re-worked to avoid code duplication.

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -85,11 +85,13 @@ Interval(interval::AbstractInterval) = convert(Interval, interval)
 Interval{T}(interval::AbstractInterval) where T = convert(Interval{T}, interval)
 
 # Endpoint constructors
-function Interval{T}(left::LeftEndpoint{T}, right::RightEndpoint{T}) where T
-    Interval{T}(left.endpoint, right.endpoint, left.included, right.included)
+function Interval{T}(left::LeftEndpoint, right::RightEndpoint) where T
+    Interval{T}(T(left.endpoint), T(right.endpoint), left.included, right.included)
 end
 
-Interval(left::LeftEndpoint{T}, right::RightEndpoint{T}) where T = Interval{T}(left, right)
+function Interval(left::LeftEndpoint{S}, right::RightEndpoint{T}) where {S, T}
+    Interval{promote_type(S, T)}(left, right)
+end
 
 # Empty Intervals
 Interval{T}() where T = Interval{T}(zero(T), zero(T), Inclusivity(false, false))
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -279,6 +279,14 @@ function Base.intersect(a::AbstractInterval{T}, b::AbstractInterval{T}) where T
     return Interval{T}(left, right)
 end
 
+function Base.intersect(a::AbstractInterval{S}, b::AbstractInterval{T}) where {S, T}
+    !overlaps(a, b) && return Interval{promote_type(S, T)}()
+    left = max(LeftEndpoint(a), LeftEndpoint(b))
+    right = min(RightEndpoint(a), RightEndpoint(b))
+
+    return Interval(left, right)
+end
+
 # There is power in a union.
 """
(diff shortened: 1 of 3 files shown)


test/interval.jl
```diff
@@ -136,46 +150,56 @@
             for i in 0:3
                 interval = Interval(a, b, Inclusivity(i))
                 cp = copy(interval)
-                lesser_val = Interval(a - unit, b + unit, Inclusivity(i))
+                lesser_val = Interval(a - unit, b - unit, Inclusivity(i))
```
omus, line 153: Change isn't required but I think it makes it clearer that this interval is always less than the original

src/interval.jl
```diff
@@ -277,6 +279,14 @@ function Base.intersect(a::AbstractInterval{T}, b::AbstractInterval{T}) where T
     return Interval{T}(left, right)
 end
 
+function Base.intersect(a::AbstractInterval{S}, b::AbstractInterval{T}) where {S, T}
+    !overlaps(a, b) && return Interval{promote_type(S, T)}()
+    left = max(LeftEndpoint(a), LeftEndpoint(b))
+    right = min(RightEndpoint(a), RightEndpoint(b))
+
+    return Interval(left, right)
+end
```
omus, line 288: I'll note this implementation could probably replace the `intersect(::AbstractInterval{T}, ::AbstractInterval{T})` defined above this

omus: As there is no constraint between them they can be anything including the same. Since there a more specific definition where they have to be the same then this method will only be called when the are different.

test/interval.jl
```diff
@@ -550,11 +575,11 @@
             Interval(-100, -1, Inclusivity(false, false)),
             Interval(-10, -1, Inclusivity(false, false)),
             Interval(10, 15, Inclusivity(false, false)),
-            Interval(13, 20, Inclusivity(false, false))
+            Interval(13, 20, Inclusivity(false, false)),
```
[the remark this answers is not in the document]
omus: Union test for infinity are included in the comparison tests

test/interval.jl
```diff
@@ -1,10 +1,20 @@
+Base.isinf(x::Char) = false
+Base.isinf(x::TimeType) = false
```
nickrobinson251, line 2: piracy?

nickrobinson251: could there at least be a comment?

omus: Definitely, it's just for tests though so it's not a big deal

nickrobinson251: should we just define and use our own `_isinf`?

omus: I may open a PR to get `isinf(::TimeType)` implemented but I very much doubt I would get a PR for `isinf(::Char)` accepted

omus: If this really bothers people I can do:
```julia
_isinf(x) = isinf(x)
_isinf(::Char) = false
_isinf(::TimeType) = false
```

nickrobinson251: i'd slightly prefer that as it make the code easier to follow (to my eyes) as it make clearly we've added some particulars for our use-case

omus: How about another compromise?
```julia
isinf(x) = Base.isinf(x)
isinf(::Char) = false
isinf(::TimeType) = false
```

omus: Something similar: https://github.com/invenia/Intervals.jl/pull/104/files#r444362741. We could actually have a generic `isinf` implemented as:
```julia
isinf(x) = !isfinite(x) && !isnan(x)
isfinite(x) = iszero(x - x)
isnan(x) = x != x
```
The only issue I see with this is that the generic `isnan` may not always hold for all types

Issue #98, opened 2020-06-03
Reduce interval storage size


omus:
This is an idea I've had for a while on but haven't investigated. Currently `Inclusivity` is defined as:
```julia
struct Inclusivity
    first::Bool
    last::Bool
end
```
This structure uses 2-bytes to store these two booleans but we can theoretically reduce the size down to 2-bits. Unfortunately Julia only allows structures as small as 8-bits/1-byte but this is still an improvement. Potentially this could save a bunch of memory when it comes to storing arrays of `Interval`s.

omus:
I did an experiment with the current version and this new version of `Inclusivity`:
```julia
struct Inclusivity
    val::UInt8

    Inclusivity(x::UInt8) = new(x & 0b11)
end


Inclusivity(x::Integer) = Inclusivity(trunc(UInt8, x))
Inclusivity(first::Bool, last::Bool) = Inclusivity(UInt8(first) | UInt8(last) << 1)

Base.copy(inc::Inclusivity) = Inclusivity(inc.val)
Base.broadcastable(inc::Inclusivity) = Ref(inc)

function Base.convert(::Type{T}, inc::Inclusivity) where T <: Integer
    return convert(T, getfield(inc, :val))
end

Base.first(inc::Inclusivity) = getfield(inc, :val) & 0b01 > 0
Base.last(inc::Inclusivity) = getfield(inc, :val) & 0b10 > 0

function Base.getproperty(inc::Inclusivity, f::Symbol)
    if f === :first
        first(inc)
    elseif f === :last
        last(inc)
    else
        getfield(inc, f)
    end
end

isclosed(inc::Inclusivity) = x == 0b11
isopen(inc::Inclusivity) = x == 0b00
```
The experiment just checked the total size of the `Interval` using the original and modified `Inclusivity`.

Original:
```julia
julia> function demo(T)
           x = Interval(T(1), T(2), true, false)
           @show T sizeof(T) sizeof(x)
       end
demo (generic function with 1 method)

julia> demo.(Base.BitInteger_types);
T = Int8
sizeof(T) = 1
sizeof(x) = 4
T = Int16
sizeof(T) = 2
sizeof(x) = 6
T = Int32
sizeof(T) = 4
sizeof(x) = 12
T = Int64
sizeof(T) = 8
sizeof(x) = 24
T = Int128
sizeof(T) = 16
sizeof(x) = 40
T = UInt8
sizeof(T) = 1
sizeof(x) = 4
[...]

PR #102, opened 2020-06-10
Encode bound information as type parameter

omus:
When updating the Intervals.jl to support unbounded ranges I encoutered an issue where the interface we had made it difficult to define unbounded ranges. For example if we update `Inclusivity` to have unbounded fields then constructor would take 4 bools: left open/closed, right open/closed, left unbounded, right unbounded:
```julia
Interval(1, 2, Inclusivity(false, true, true, false))  # `(,2]`
```
(Aside: note we still store a value for the unbounded side as we want `Interval` to be a bits type. The value here ultimately is ignored as it is unbounded)

This interface isn't very intuitive and needs some revision. While looking into [reducing storage size of intervals](https://github.com/invenia/Intervals.jl/issues/98) I discovered that we can reduce the size of an interval in memory by having the bound information encoded in the the type parameters. The clearest way to do this is to use symbols. Using the same example as above we could have:
```julia
Interval{:unbounded, :closed}(1, 2)
```
With the option to alternatively construct the same interval with in the future:
```julia
Interval(nothing, 2)
```

This PR currently implements minimal working changes that support encoding bound information as type parameters. At this point there is no support for unbounded ranges. If we like this interface I can update the remainder of the code. If we don't like this I could work on an alternative interface such as:
```julia
Interval(1, 2, :unbounded, :closed)
```


src/Intervals.jl
```diff
@@ -9,7 +9,13 @@ using Dates: AbstractDateTime, value, coarserperiod
 
 import Base: ⊆, ⊇, ⊈, ⊉, union, union!, merge
 
-abstract type AbstractInterval{T} end
+abstract type Bound end
+struct Closed <: Bound end
+struct Open <: Bound end
+
+bound(x::Bool) = x ? Closed : Open
```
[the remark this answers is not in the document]
omus: I went with `bound_type` as we also have `bounds_types` and I didn't like `boundstypes`

omus:
PR should now be complete and can be reviewed at this time. I'll call out the couple of minor negative points to do with this change:
1. `Interval{Open, Closed}(1, 2)` works as you'd expect but `Interval{Open, Closed}` won't get you the type you expect. Instead you need to write `Interval{T,Open,Closed} where T`
2. The `HE` and `HB` constructors have no way of supplying an alternative bound types. I don't believe this an issue and `HourEnding{L,R}(...)` and `HourBeginning{L, R}(...)` works as you expect (Note: the `HE` and `HB` perform `ceil`/`floor` to the nearest hour)

src/anchoredinterval.jl
```diff
@@ -136,37 +135,37 @@ span(interval::AnchoredInterval{P}) where P = abs(P)
...
 
-function Base.convert(::Type{Interval}, interval::AnchoredInterval{P, T}) where {P, T}
-    return Interval{T}(first(interval), last(interval), inclusivity(interval))
+function Base.convert(::Type{Interval}, interval::AnchoredInterval{P,T,L,R}) where {P,T,L,R}
+    return Interval{T,L,R}(first(interval), last(interval))
 end
 
-function Base.convert(::Type{Interval{T}}, interval::AnchoredInterval{P, T}) where {P, T}
-    return Interval{T}(first(interval), last(interval), inclusivity(interval))
+function Base.convert(::Type{Interval{T}}, interval::AnchoredInterval{P,T,L,R}) where {P,T,L,R}
```
fchorney, line 142: Is there a situation where this function is used over the one above? or more so, would there ever be a difference?

src/anchoredinterval.jl
```diff
@@ -194,57 +191,60 @@ end
...
 
 ##### SET OPERATIONS #####
 
-function Base.isempty(interval::AnchoredInterval{P, T}) where {P, T}
+function Base.isempty(interval::AnchoredInterval{P,T}) where {P,T}
     return P == zero(P) && !isclosed(interval)
 end
 
-function Base.intersect(a::AnchoredInterval{P, T}, b::AnchoredInterval{Q, T}) where {P,Q,T}
+function Base.intersect(a::AnchoredInterval{P,T}, b::AnchoredInterval{Q, T}) where {P,Q,T}
```
fchorney, line 247:
```suggestion
function Base.intersect(a::AnchoredInterval{P,T}, b::AnchoredInterval{Q,T}) where {P,Q,T}
```

src/description.jl
```diff
@@ -1,19 +1,19 @@
...
-        first(inclusivity(interval)) ? '[' : '(',
+        L === Closed ? '[' : '(',
         description(anchor(interval), abs(P), s),
-        last(inclusivity(interval)) ? ']' : ')',
+        R === Closed ? ']' : ')',
     )
 end
 
-function description(interval::AnchoredInterval{P, ZonedDateTime}, s::String) where P
+function description(interval::AnchoredInterval{P, ZonedDateTime, L, R}, s::String) where {P,L,R}
```
fchorney, line 11: Do we only remove spaces after commas in type lists when all the types are single chars?

nickrobinson251:
> The HE and HB constructors have no way of supplying an alternative bound types

🤷 Different issue, but `HE` and `HB` are the least nice interface in this package already #105

src/description.jl
omus: Yes, for "simple" parametric types I've removed the spaces as otherwise some lines would need to be broken up more. It's not a standard convention of ours but it works pretty well for this package.

src/anchoredinterval.jl
omus: The signatures of these two functions is identical with the updated `AnchoredInterval` type.

omus:
Just applied Fernando's change and corrected the doctests

src/anchoredinterval.jl
fchorney: Do we need them both then?

omus: I misinterpreted your original question. The actual answer to your initial question is that there is a difference between the two functions. One is called as `convert(Interval, ...)` which is just "convert this to a subtype of Interval" where `convert(Interval{Int64}, ...)` is "convert this to a subtype of `Interval` where the element type is `Int64`".

We could combine these so the signature is:
```
Base.convert(::Union{Type{Interval}, Type{Interval{T}}, ...)
```
I may make that change in another PR.

fchorney: I think this is just me trying to wrap my head around multiple dispatch for certain things. I get that one is just strictly `Interval` and the other is `Interval{T}`, but doesn't the `Interval{T}` one stipulate that the `T` has to be the same type as the `T` in the `AnchoredInterval{P,T,L,R}`?, and therefore you wouldn't be able to do something like `convert(Interval{Int64})` if the `T` in the `AnchoredInterval` was a `Float64` per se? Perhaps I am misunderstanding, and both `T`'s don't actually need to match?

PR #104, opened 2020-06-10
Unbounded interval support

omus:
Using the parametric bounds interface in https://github.com/invenia/Intervals.jl/pull/102 this PR adds the fundamentals for constructing and interacting with unbounded intervals.

Intervals with unbounded endpoints can be constructed by using `nothing` as a placeholder (e.g. `Interval(nothing, 2)` creates the interval `(,2]`). Explicitly stating the bound type requires that when supplying `nothing` that the respectively stated bound type is `Unbounded` (e.g. `Interval{Unbounded,Open}(nothing, 3)` creates `(,3)` but `Interval{Open,Open}(nothing, 3)` and `Interval{Unbounded,Unbounded}(nothing, 3)` both raise `MethodError`s).

As for the actual implementation for `Interval{T}` where `T` is the element type we always store values of type `T` and never store `Nothing`. Doing this avoids using a union type which would result in `Interval` no longer being a bits type. However, using the `first` or `last` function on an unbounded endpoint does now return `nothing`.

The current implementation has `span` throw an exception on unbounded intervals as discussed in https://github.com/invenia/Intervals.jl/issues/89#issuecomment-638918630.


src/endpoint.jl
```diff
@@ -82,26 +108,65 @@ function Base.isequal(a::RightEndpoint, b::LeftEndpoint)
 end
 
 function Base.isless(a::LeftEndpoint, b::LeftEndpoint)
-    a.endpoint < b.endpoint || (a.endpoint == b.endpoint && isclosed(a) && !isclosed(b))
+    return (
+        !isunbounded(b) && (
```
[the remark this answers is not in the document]
omus: I think using `isunbounded` consistently in these `Endpoint` comparison functions ends up being the most readable. Definitely is just a personal opinion though.

I should mention `isunbounded(::AbstractInterval)` and `isbounded(::AbstractInterval)` may surprise you since: `isunbounded(interval) != !isbounded(interval)`. This is due to `isunbounded` being defined as both endpoints are unbounded while `isbounded` is defined as both endpoints being bounded. This ends up being useful as `!isbounded(interval)` means at least one endpoint is unbounded.

omus:
Rebased now that https://github.com/invenia/Intervals.jl/pull/102 has been merged.

src/anchoredinterval.jl
```diff
@@ -122,16 +133,31 @@ end
...
 end
 
 anchor(interval::AnchoredInterval) = interval.anchor
-span(interval::AnchoredInterval{P}) where P = abs(P)
+
+function span(interval::AnchoredInterval{P}) where P
+    if !isunbounded(interval)
+        abs(P)
+    else
+        throw(ArgumentError("Unable to determine the span of an unbounded interval"))
```
omus, line 158: Maybe should be a `DomainError`?

src/Intervals.jl
```diff
@@ -10,8 +10,10 @@ using Dates: AbstractDateTime, value, coarserperiod
 import Base: ⊆, ⊇, ⊈, ⊉, union, union!, merge
 
 abstract type Bound end
-struct Closed <: Bound end
-struct Open <: Bound end
+abstract type Bounded <: Bound end
+struct Closed <: Bounded end
+struct Open <: Bounded end
+struct Unbounded <: Bound end
```
nicoleepp, line 16: Maybe would be good to have a brief explanation in the README

src/anchoredinterval.jl
nicoleepp: Could be either I think

omus: I've updated to use a `DomainError`

src/anchoredinterval.jl
```diff
@@ -160,12 +186,25 @@ function Base.convert(::Type{AnchoredInterval{P}}, interval::Interval{T}) where
 end
 =#
 
-function Base.convert(::Type{AnchoredInterval{Ending}}, interval::Interval{T,L,R}) where {T,L,R}
-    AnchoredInterval{-span(interval), T, L, R}(last(interval))
+one(T::Type{<:TimeType}) = eps(T)
```
nickrobinson251, line 189: this seems highly confusing.

If i see `one` in code, i expect it to be `Base.one` (or have the same semantics) which is to be the multiplicative identity (which `eps(T)` is not)

omus: I renamed this to `_span_fallback`

src/Intervals.jl
omus: Added in https://github.com/invenia/Intervals.jl/pull/104/commits/c40d64ce3693aada0ae465fb39a008dfc8931cb0

src/endpoint.jl
```diff
@@ -21,18 +30,29 @@ const RightEndpoint{T,B} = Endpoint{T, Right, B} where {T,B}
...
     h = hash(D, h)
     h = hash(B, h)
-    h = hash(x.endpoint, h)
+
+    # Unbounded endpoints should ignore value.
+    if B !== Unbounded
+        h = hash(x.endpoint, h)
+    else
+        h = hash(T, h)
+    end
```
nickrobinson251, line 54: could be one line? `h = B !== Unbounded ? hash(x.endpoint, h) : hash(T, h)`

src/anchoredinterval.jl
omus: See https://github.com/invenia/Intervals.jl/pull/104/commits/59ce3fe32cd9227d0410cafdce72a9aa6d038a57

src/anchoredinterval.jl
```diff
@@ -161,12 +168,25 @@ function Base.convert(::Type{AnchoredInterval{P}}, interval::Interval{T}) where
 end
 =#
 
+one(T::Type{<:TimeType}) = eps(T)
+one(x) = Base.one(x)
```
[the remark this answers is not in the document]
omus: Renamed to `_span_fallback`

src/anchoredinterval.jl
nickrobinson251: thanks!

src/anchoredinterval.jl
```diff
@@ -60,6 +60,17 @@ See also: [`Interval`](@ref), [`HE`](@ref), [`HB`](@ref)
 """
 struct AnchoredInterval{P, T, L <: Bound, R <: Bound} <: AbstractInterval{T,L,R}
     anchor::T
+
+    function AnchoredInterval{P,T,L,R}(anchor::T) where {P, T, L <: Bound, R <: Bound}
+        if P < zero(P) && R === Unbounded || P > zero(P) && L === Unbounded
```
iamed2, line 65: Should we really include support for unbounded AnchoredIntervals? Do you expect people to pass infinite periods? I think it makes sense to restrict AnchoredIntervals to finite `P`.

src/anchoredinterval.jl
```diff
@@ -160,12 +186,25 @@ function Base.convert(::Type{AnchoredInterval{P}}, interval::Interval{T}) where
 end
 =#
 
-function Base.convert(::Type{AnchoredInterval{Ending}}, interval::Interval{T,L,R}) where {T,L,R}
-    AnchoredInterval{-span(interval), T, L, R}(last(interval))
+_span_fallback(::Type{T}) where T <: TimeType = eps(T)
+_span_fallback(::Type{T}) where T = one(T)
+
+function Base.convert(::Type{AnchoredInterval{Ending}}, interval::Interval{T}) where {T}
```
iamed2, line 192: I hadn't seen this before but this is a problematic usage of type parameters and should be removed, if not in this PR than another. First, `Ending` is not a period, second, you can't have an instance of this type, and third, the invariant `convert(T, x)::T` should always hold in every `convert` method.

src/endpoint.jl
omus: It can be. I thought this was easier to read but I was on the fence

src/anchoredinterval.jl
omus: A very valid question. Just because we can support it doesn't mean we should. I'll give that some thought

src/anchoredinterval.jl
omus: I can agree that this is weird but these end up being very useful (mostly for tests though). The `Ending`/`Beginning` is just used to specify what anchor should be used as otherwise we would otherwise pick one at random. This is worth making a separate issue about but I'm reluctant to make that change here.

src/anchoredinterval.jl
omus: One argument for keeping this is that removing `Unbounded` support doesn't stop people from using infinite periods. We'd have to add checks to disable that as well

omus: Thinking about this more I think support for unbounded and infinity should be consistent. If we do want to remove unbounded support entirely from `AnchoredInterval` we should also remove support for infinity.

My rational in why these go together is that I was planning on defining equality between infinite and unbounded intervals such that `[∞,x] == (,x]`. Not having unbounded support but having infinity support seems wrong to me considering this.

src/anchoredinterval.jl
iamed2: IMO that should be a constructor argument (or keyword argument), or this should just be a different function.

I understand not wanting to make the change in this PR.

src/anchoredinterval.jl
iamed2: > Thinking about this more I think support for unbounded and infinity should be consistent.

Yeah I do agree.

Okay, I can imagine wanting this rather than just using Interval I guess. But it's hard to imagine why you'd want an AnchoredInterval with infinity as part of the type.

iamed2: Okay, I'm okay with keeping this functionality. But a `MethodError` hurts readability for the error, since the user cannot see from the type signature why this is wrong. I think this should be a different type of error. 

One way to keep it a MethodError would be to add yet another type parameter that's just the sign of the period, then all the relevant information would be available for dispatch.

omus: I'll make it an argument error. Not entirely accurate but better

omus:
In reviewing the code again and taking [this conversation](https://github.com/invenia/Intervals.jl/pull/104#discussion_r439577261) into consideration. I've noticed that there are some cases with `AnchoredInterval`where non-finite values are allowed as the value for the anchor endpoint. The reason this is problematic is for the same reason we don't allow the anchor endpoint to be unbounded: it makes it impossible to calculate the other endpoint.

I've unfortunately run out of time to work on this and ensure the problem is solved correctly so I'll be holding off on merging this PR. I do believe the unbounded support for the `Interval` type is fully functional at this time.

omus:
After much consideration about non-bounded `AnchoredInterval`s I've distilled my thoughts down to the following:

With the early implementation for unbounded interval we allowed the creation of instances such as `Interval{Unbounded,Unbounded}(1, 2)`. Although the user has specified values for the lower/upper endpoint values they were ignored as the type of the interval was specified as unbounded. To avoid confusion `nothing` constructors were introduced which no longer allowed values to be specified for unbounded endpoints (even though internally values are stored in the field) and instead required `nothing` to be used where the endpoint was specified to be unbounded (e.g. `Interval{Unbounded,Unbounded}(nothing, nothing)`).

For `AnchoredInterval`s the anchor endpoint position (lower/upper) is determined by the sign of the span (+/-) and the other endpoint is computed via the `anchor + span`. Due to how an `AnchoredInterval` internal representation the anchor endpoint must be bounded in order to compute the other endpoint if it is not specified to be unbounded. Specifically, `[0,)` cannot be represented by an `AnchoredInterval` when the anchor is the upper endpoint as there is no span which allows us to compute the other endpoint of `0`. Alternatively, `[0,)` can somewhat be represented by an `AnchoredInterval` when the anchor is the lower endpoint (e.g. `AnchoredInterval{+x,Closed,Unbounded}(0)` where `x` is any positive integer). In the last example only the sign of the span matters in specifying if the anchor's endpoint position. A major caveat with this approach is that the span itself isn't actually representative of an non-bounded interval.

For non-bounded `Interval`s we decided that the computed span should raise an exception since for intervals with element types which have no representation of infinity in a type stable way. With this in mind and also considering that an non-bounded `AnchoredInterval` has the span as part of the type (meaning it easily visible) makes me think that `AnchoredInterval`s should not support non-bounded intervals.

Finally, considering intervals using infinity: since we have a way of representing infinity for certain types I see no issue with continuing to support non-finite values as long as the either endpoint and the span are not nan.

```julia
AnchoredInterval{-Inf}(0)   # (Inf,0]
AnchoredInterval{Inf}(0)    # [0,Inf)
AnchoredInterval{0}(Inf)    # [Inf,Inf]
AnchoredInterval{-10}(Inf)  # [Inf,Inf]
AnchoredInterval{-Inf}(Inf) # Invalid as `Inf + (-Inf) = NaN`
```

TLDR; `AnchoredInterval` should always use a span value which represents the actual span of the interval. As non-bounded intervals have no defined span we will not support non-bounded `AnchoredInterval`s. However, infinite anchored intervals will still be supported.

src/anchoredinterval.jl
omus: More thoughts on this: https://github.com/invenia/Intervals.jl/pull/104#issuecomment-648245619

test/interval.jl
```diff
@@ -42,16 +44,28 @@ isinf(::TimeType) = false
...
+        @test_throws MethodError Interval{Int, Closed, Unbounded}(1, 2)
+        @test_throws MethodError Interval{Int, Unbounded, Unbounded}(1, 2)
     end
 
     @testset "non-ordered" begin
-        @test_throws ArgumentError Interval(NaN, Inf)
         @test_throws ArgumentError Interval(NaN, NaN)
+        @test_throws ArgumentError Interval(NaN, Inf)
+        @test_throws ArgumentError Interval(-Inf, NaN)
+        # @test_throws ArgumentError Interval(Inf, -Inf)  # Would result in a NaN span
```
omus, line 68: Can't currently be tested due to automatic endpoint reordering.

omus: Once [this deprecation](https://github.com/invenia/Intervals.jl/pull/97) is added and finally dropped this test can be enabled.

src/anchoredinterval.jl
omus: https://github.com/invenia/Intervals.jl/issues/109

src/anchoredinterval.jl
```diff
@@ -58,8 +58,41 @@ AnchoredInterval{5 minutes,DateTime,Closed,Closed}(2016-08-11T12:30:00)
...
+                throw(ArgumentError(msg))
+            end
+        end
+
+        return new{P,T,L,R}(anchor)
+    end
+end
+
+_isfinite(x) = iszero(x - x)
+_isfinite(x::Real) = Base.isfinite(x)
```
omus, line 92: I'm attempting to get this functionality added to base: https://github.com/JuliaLang/julia/pull/36380

src/docstrings.jl
```diff
@@ -0,0 +1,107 @@
...
+
+Type indicating that the endpoint of an interval is open (the endpoint value is *not
+included* in the interval).
+"""
+:Open
+
+"""
+    Unbounded <: Bound
+
+Type indicating that the endpoint of an interval is unbounded (the endpoint value
```
TylerLoewen, line 39:
```suggestion
Type indicating that the endpoint of an interval is unbounded (the endpoint value is
```

src/docstrings.jl
```diff
@@ -0,0 +1,107 @@
...
+
+The delta between the upper and lower endpoints. For bounded intervals returns a
+non-negative value while intervals with any unbounded endpoints will throw an
+`ArgumentError`.
+
+To avoid having to capture the exception use the pattern:
+```julia
+!isbounded(interval) ? span(interval) : infinity`
+```
+Where `infinity` is a variable representing the value you which to use to represent an
```
TylerLoewen, line 72:
```suggestion
Where `infinity` is a variable representing the value you whish to use to represent an
```

src/plotting.jl
```diff
@@ -3,7 +3,8 @@ interval_markers(L::Type{<:Bound}, R::Type{<:Bound}) = interval_marker.([L, R])
 interval_marker(x::Type{Closed}) = :vline
 interval_marker(x::Type{Open}) = :none
 
-@recipe function f(xs::AbstractVector{<:AbstractInterval{T}}, ys) where T
+# TODO: Add support for plotting unbounded intervals
```
nicoleepp, line 6: Open an issue and add to this comment?

omus: Yes, thanks. I did look into this and it got complicated quickly

omus: https://github.com/invenia/Intervals.jl/issues/110

omus:
Rebased to apply documentation fixups and fixed typo in commit message. Just waiting on CI then will merge.

Issue #90, opened 2020-05-21
Deprecate `first`/`last` in favor of `lower`/`upper`


omus:
It can be rather annoying to read code with a `Vector{Interval}` as you can end up with things like: `first(first(v))` which would be more readable as `lower(first(v))`

omus:
I'll also note this is the terminology used by PostgreSQL's range functions: https://www.postgresql.org/docs/9.3/functions-range.html

nickrobinson251:
sure, seems fine to deprecate`first`,`last`; they are somewhat associated with iteration too, i.e. it's usual to expect `first(x) == x[1]`, `last(x) == x[end]` which of course is not true of non-iterable intervals. `lower` and `upper` seem as good a name as anything else.

nickrobinson251:
a more verbose option could be `lowerbound`, `upperbound`

glennmoy:
> Although, if using `minumum`/`maximum`, then i think we'd want to respect the `inclusivity` such that we always had e.g. `minimum(x) in x` (this is not true for the current `first`/`last`). So perhaps `minimum`/`maximum` should be the inclusivity-respecting version of `lower[bound]`/`upper[bound]`... but that's now off-topic.

To stray further...the [`infimum` and `supremum`](https://en.wikipedia.org/wiki/Infimum_and_supremum) could/would supplant `first` and `last` if we wanted to keep this functionality but use more mathematically appropriate terms.

nickrobinson251:
- `infimum` and `supremum` names match [IntervalSets](https://github.com/JuliaMath/IntervalSets.jl/blob/162bd0d9ae952347980a3e2f7e6561e5170806bb/src/IntervalSets.jl#L19), [DomainSets.jl](https://github.com/JuliaApproximation/DomainSets.jl/blob/3bc85eef1c9676c662a78683aeedd34d8a8e59a0/src/generic/domain.jl#L120-L121), and [TensorKit.jl](https://github.com/Jutho/TensorKit.jl/blob/911254ba0fba9037c0968f37f659f0a26dba51a9/src/spaces/cartesianspace.jl#L36-L37) (i _think_ these all have pretty consistent semantics even though they don't extend a common function)
- `inf` and `sup` names matches [InveralArithmetic.jl](https://github.com/JuliaIntervals/IntervalArithmetic.jl/blob/a48c26a837fb7e8e00138cef75064061551ae59d/src/intervals/arithmetic.jl#L304-L306)

[...]

PR #108, opened 2020-06-19
Additional accessor and convert functions

sprockmonty:
Added some functions simple `containsfirst()` and `constainslast()` functions and a convert function to allow Intervals to be added to Interval arrays with different types e.g.
```julia
x = Interval{Int,Closed, Open}(1,2)
containsfirst(x) # == true
containslast(x) # == false
y = Interval{Float64,Closed,Open}(1.0,2.0)
a = [x,x]
push!(a,y)
```


src/interval.jl
```diff
@@ -137,6 +139,10 @@ function Base.convert(::Type{T}, i::Interval{T}) where T
     throw(DomainError(i, "The interval is not closed with coinciding endpoints"))
 end
 
+function Base.convert(::Type{Interval{T,L,R}}, interval::Interval{N,L,R}) where {T,N,L,R}
+    return Interval{L,R}(convert(T, first(interval)), convert(T, last(interval)))
```
omus, line 143:
```suggestion
    return Interval{T,L,R}(first(interval), last(interval))
```

src/interval.jl
```diff
@@ -120,6 +120,8 @@ Base.first(interval::Interval) = interval.first
 Base.last(interval::Interval) = interval.last
 isclosed(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Closed && R === Closed
 Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R === Open
+containsfirst(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Closed 
+containslast(interval::AbstractInterval{T,L,R}) where {T,L,R} = R === Closed
```
omus, line 124: I think this functionality is useful but the name needs more consideration. There may be a rename of first/last in the near future: https://github.com/invenia/Intervals.jl/issues/90

sprockmonty:
I've made the change and added an additional promote_rule function for promoting intervals (in the hope of replacing the second intersect function but it's not quite there yet).

src/interval.jl
```diff
@@ -189,6 +189,14 @@ function Base.convert(::Type{T}, interval::Interval{T}) where T
     end
 end
 
+function Base.convert(::Type{<:Interval{T}}, interval::Interval{S,L,R}) where {T,S,L,R}
```
omus, line 192:
```suggestion
function Base.convert(::Type{Interval{T}}, interval::Interval{S,L,R}) where {T,S,L,R}
```

test/interval.jl
```diff
@@ -85,6 +85,7 @@ isinf(::TimeType) = false
         @test_throws DomainError convert(Int, Interval{Closed, Open}(10, 10))
         @test convert(Int, Interval{Closed, Closed}(10, 10)) == 10
         @test_throws DomainError convert(Int, Interval{Closed, Closed}(10, 11))
+        @test convert(Interval{Float64, Closed, Closed}, Interval(1,2)) == Interval{Float64, Closed, Closed}(1.0,2.0)
```
omus, line 88: Hmmm, `convert(Interval{Float64, Open, Open}, Interval(1,2))` isn't lossless. This means that the convert method should probably be defined as: `Base.convert(::Type{Interval{T}}, interval::Interval{S,L,R}) where {T,S,L,R}`

test/interval.jl
```diff
@@ -101,6 +102,20 @@ isinf(::TimeType) = false
...
     end
 
+    @testset "promotion" begin
+        for (a1, b1, _) in test_values
+            for (a2, b2, _) in test_values
+                for (L1, R1) in BOUND_PERMUTATIONS
+                    for (L2, R2) in BOUND_PERMUTATIONS
+                        interval1 = Interval{L1, R1}(a1, b1)
+                        interval2 = Interval{L2, R2}(a2, b2)
+                        @test promote(interval1, interval2) == (Interval{L1,R1}(promote(a1, b1)...), Interval{L2,R2}(promote(a2, b2)...))
```
omus, line 112:
```suggestion
                        interval1 = Interval{L1, R1}(a1, b1)
                        interval2 = Interval{L2, R2}(a2, b2)
                        promoted1 = Interval{L1,R1}(promote(a1, b1)...)
                        promoted2 = Interval{L2,R2}(promote(a2, b2)...)
                        @test promote(interval1, interval2) == (promoted1, promoted2)
```

src/interval.jl
```diff
@@ -189,6 +189,14 @@ function Base.convert(::Type{T}, interval::Interval{T}) where T
...
+    return Interval{T,L,R}(first(interval), last(interval))
+end
+
+Base.promote_rule(::Type{Interval{T,L1,R1}}, ::Type{Interval{S,L2,R2}}) where {T,S,L1,R1,L2,R2} = Interval{promote_type(T,S), <:Union{L1,L2}, <:Union{R1,R2}}
+# hacky way to fix situations where T and S are the same but L and R are different
+Base.not_sametype(x::Tuple{Interval{T,L1,R1}, Interval{T,L2,R2}}, y::Tuple{Interval{T,L1,R1}, Interval{T,L2,R2}}) where {T,L1,R1,L2,R2} = nothing
```
omus, line 198: Probably we should just define:
```julia
Base.promote_rule(::Type{Interval{T,L,R}}, ::Type{Interval{S,L,R}}) where {T,S,L,R} = Interval{promote_type(T,S), L, R}
```
I suspect this would eliminate the need for the hack

test/interval.jl
sprockmonty: Would it also be useful to include `Base.convert(::Type{Interval{T,L,R}}, interval::Interval{S}) where {T,S,L,R}` such that if you need to convert from Closed Closed to e.g. Open Open you can? Should this type of conversion include Closed -> Open but not Open -> Closed

omus: I think this would make more sense as a constructor; https://docs.julialang.org/en/v1/manual/conversion-and-promotion/#Conversion-vs.-Construction-1.

I'd probably leave this out unless you have a use case.

PR #111, opened 2020-06-23
Fix AnchoredInterval scalar convert instead of deprecating

omus:
A continuation of https://github.com/invenia/Intervals.jl/pull/101. In that PR I deprecated `convert(::Type{T}, ::AnchoredInterval{P,T})` as the conversion was lossy. Recently I noticed `convert(::Type{T}, ::Interval{T})` which converts intervals with only a single element into a scalar. It appears that the `AnchoredInterval` version of this function was just implemented incorrectly.

Testing of deprecations:
```julia
julia> using Intervals, Dates

julia> Date(Interval(today(), today()))
┌ Warning: `Date(interval::Interval{Date})` is deprecated, use `convert(Date, interval)` instead.
│   caller = ip:0x0
└ @ Core :-1
2020-06-23

julia> DateTime(Interval(now(), now()))
┌ Warning: `DateTime(interval::Interval{DateTime})` is deprecated, use `convert(DateTime, interval)` instead.
│   caller = ip:0x0
└ @ Core :-1
2020-06-23T14:38:18.26

julia> Date(AnchoredInterval{Hour(0)}(today()))
┌ Warning: `Date(interval::(AnchoredInterval{P, Date} where P))` is deprecated, use `convert(Date, interval)` instead.
│   caller = ip:0x0
└ @ Core :-1
2020-06-23

julia> DateTime(AnchoredInterval{Hour(0)}(now()))
┌ Warning: `DateTime(interval::(AnchoredInterval{P, DateTime} where P))` is deprecated, use `convert(DateTime, interval)` instead.
│   caller = ip:0x0
└ @ Core :-1
2020-06-23T14:38:39.605
```

I'll note we could probably remove these scalar convert methods as they have limited use but it's probably best to correct the function first.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -6,11 +6,6 @@ import Dates: Date, DateTime
 export Inclusivity, inclusivity
 include("inclusivity.jl")
 
-function Base.convert(::Type{T}, interval::AnchoredInterval{P,T}) where {P,T}
-    depwarn("`convert($T, interval::AnchoredInterval{P,$T})` is deprecated, use `anchor(interval)` instead.", :convert)
-    anchor(interval)
-end
-
 @deprecate Date(interval::Interval{Date}) convert(Date, interval)
 @deprecate DateTime(interval::Interval{DateTime}) convert(DateTime, interval)
 @deprecate Date(interval::AnchoredInterval{P, Date} where P) convert(Date, interval)
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -1,4 +1,5 @@
-using Base: depwarn
+using Base: @deprecate, depwarn
+import Dates: Date, DateTime
 
 # BEGIN Intervals 1.X.Y deprecations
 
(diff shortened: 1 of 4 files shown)


src/deprecated.jl
```diff
@@ -1,21 +1,15 @@
...
 
 # BEGIN Intervals 1.X.Y deprecations
 
 export Inclusivity, inclusivity
 include("inclusivity.jl")
 
-function Base.convert(::Type{T}, interval::AnchoredInterval{P,T}) where {P,T}
-    depwarn("`convert($T, interval::AnchoredInterval{P,$T})` is deprecated, use `anchor(interval)` instead.", :convert)
-    anchor(interval)
-end
```
omus, line 11: Note we haven't had a tag since the deprecation was added.

src/deprecated.jl
```diff
@@ -1,21 +1,15 @@
-using Base: depwarn
+using Base: @deprecate, depwarn
```
nicoleepp, line 1: Is depwarn actually used anywhere?

omus: Yes, it's used later in this file

PR #113, opened 2020-06-24
Add deprecation for old scalar convert

omus:
I overlooked in https://github.com/invenia/Intervals.jl/pull/111 that I needed to have a deprecation when conversion was invalid.

Testing of deprecations
```julia
julia> using Intervals, Dates

julia> convert(DateTime, AnchoredInterval{Hour(0)}(DateTime(2000)))
2000-01-01T00:00:00

julia> convert(DateTime, HourEnding(DateTime(2000)))
┌ Warning: `convert(T, interval::AnchoredInterval{P,T})` is deprecated for intervals which are not closed with coinciding endpoints. Use `anchor(interval)` instead.
│   caller = top-level scope at REPL[3]:1
└ @ Core REPL[3]:1
2000-01-01T00:00:00
```

### DIFF
--- a/test/anchoredinterval.jl
+++ b/test/anchoredinterval.jl
@@ -107,8 +107,14 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
 
         he = HourEnding(dt)
         hb = HourBeginning(dt)
-        @test_throws DomainError convert(DateTime, he)
-        @test_throws DomainError convert(DateTime, hb)
+
+        # Note: When the deprecation is dropped remove the deprecated tests and uncomment
+        # the DomainError tests
+        @test (@test_deprecated convert(DateTime, he)) == anchor(he)
+        @test (@test_deprecated convert(DateTime, hb)) == anchor(hb)
+        # @test_throws DomainError convert(DateTime, he)
+        # @test_throws DomainError convert(DateTime, hb)
+
         @test convert(Interval, he) == Interval{Open, Closed}(dt - Hour(1), dt)
         @test convert(Interval, hb) == Interval{Closed, Open}(dt, dt + Hour(1))
         @test convert(Interval, he) == Interval{Open, Closed}(dt - Hour(1), dt)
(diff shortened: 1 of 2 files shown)


mjram0s:
Do any of the tests for `convert(T, interval::AnchoredInterval{P,T})` need to be updated to reflect this?

omus:
> Do any of the tests for `convert(T, interval::AnchoredInterval{P,T})` need to be updated to reflect this?

Yes, I forgot I added some tests which expect to get a `DomainError`

omus:
Added `@test_broken` tests. Unfortunately there isn't a `@test_broken_throws` so we have something a little more messy

omus:
Thanks to @morris25 for noticing this

test/anchoredinterval.jl
```diff
@@ -107,8 +107,27 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
 
         he = HourEnding(dt)
         hb = HourBeginning(dt)
-        @test_throws DomainError convert(DateTime, he)
-        @test_throws DomainError convert(DateTime, hb)
+
+        # Note: replace try/catch with the following when deprecation is dropped
+        # @test_throws DomainError convert(DateTime, he)
+        @test_broken try
```
morris25, line 113: Is `@test_broken` the right thing to use here? Is this expected behavior for this version (that will soon change) or just a bug that nobody noticed?

omus: I think this is definitely debatable. Specifically the tests which expect a `DomainError` are broken. However, for this version we do expect that the convert will pass so I could use a `@test` here. When the deprecation is dropped that test will fail and result in the tests also being updated which is really what I'm after.

omus: I'll change this as the current implementation is probably obfuscating things

morris25: You could just use `@test_deprecated` which will fail when the code changes

PR #115, opened 2020-06-25
Appveyor only tests plotting

omus:
I've updated the CI testing so that Appveyor only does plotting testing on Windows and Travis CI does the remainder of the Windows tests. Ideally this should decrease overall CI time. Note there is currently a bug with FFMPEG.jl and Trvais CI which means we cannot test plotting on Windows in Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,11 @@
 language: julia
 julia:
   - 1.0  # Latest LTS release
-  - 1  # Latest stable release
+  - 1    # Latest stable release
   - nightly
 os:
   - linux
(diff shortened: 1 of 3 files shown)


omus:
Looks like I have a CI issue to sort out and I need to report on the speed improvements

pabloferz:
This is obsolete, right?

omus:
> This is obsolete, right?

Yes, this is pretty outdated but limiting the where we test plotting is still a consideration. I'll try and validate the performance is bad enough that this kind of change is still needed.

PR #119, opened 2020-06-26
Use `@test_deprecated` on Inclusivity

omus:
Verifies that `Inclusivity` is deprecated and more importantly silences the deprecation warnings.

I may have noticed a bug with this and using `--depwarn=error` which I'll investigate separately.

### DIFF
--- a/test/inclusivity.jl
+++ b/test/inclusivity.jl
@@ -1,3 +1,5 @@
+Inclusivity(args...) = @test_deprecated Intervals.Inclusivity(args...)
+
 @testset "Inclusivity" begin
     @testset "constructor" begin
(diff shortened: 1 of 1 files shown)


test/inclusivity.jl
```diff
@@ -1,3 +1,5 @@
+Inclusivity(args...) = @test_deprecated Intervals.Inclusivity(args...)
```
nicoleepp, line 1: Why define this function, doesn't seem to be used?

omus: There are many calls to `Inclusivity` in this file. I did this to avoid having to do `(@test_deprecated Inclusivity(...))` for each one

Issue #121, opened 2020-06-29
Image not visible in Plotting Section : Docs


LakshyaKhatri:
In [Plotting](https://invenia.github.io/Intervals.jl/stable/#Plotting-1) section of the documentation, there is supposed to be an image here instead of the word `plot`:

![Screenshot from 2020-06-29 11-46-36](https://user-images.githubusercontent.com/28972442/85979487-37f73100-b9fe-11ea-8936-309d9f8d8212.png)

LakshyaKhatri:
Also, there are some more issues. Kindly let me know if you are open to contributions to the core library. Thanks :smile:

omus:
Thanks for reporting this. I suspect when we updated Documenter with version 1.3.0 this started happening. I'll look into this today

LakshyaKhatri:
Thanks for your quick response @omus. Also, in the API section [(Intervals.span — Function)](https://invenia.github.io/Intervals.jl/latest/#Intervals.span) I have two doubts:
 1. The `isbounded` and `isunbounded` functions are not exported. Is that the expected behaviour? If so, it will be helpful to mention that in the docs.
 2. Is that trailing **( ` )** intentional in the second snippet?
![Screenshot from 2020-06-29 22-53-41](https://user-images.githubusercontent.com/28972442/86037043-6d326c00-ba5c-11ea-851e-4716c54725a9.png)

LakshyaKhatri:
cc: @omus

PR #124, opened 2020-06-29
Fix the plotting example in documentation

omus:
I believe the plot only ever rendered correctly when viewing from GitHub and not the hosted documentation. 

Fixes: https://github.com/invenia/Intervals.jl/issues/121


mattBrzezinski:
Is this file generated? I'm not seeing existing in the [assets dir](https://github.com/invenia/Intervals.jl/tree/master/docs/src/assets)

omus:
The file is added by this [commit](https://github.com/invenia/Intervals.jl/tree/cv/docs-plot-fix/docs/src/assets). The link you posted is for the master branch

Issue #125, opened 2020-06-30
isbounded() and isunbounded() are not exported


LakshyaKhatri:
Thanks @omus for a quick response on the previous issue. Also, in the API section [(Intervals.span — Function)](https://invenia.github.io/Intervals.jl/latest/#Intervals.span) I have two doubts:
 1. The `isbounded` and `isunbounded` functions are not exported. Is that the expected behaviour? If so, it will be helpful to mention that in the docs.
 2. Is that trailing **( ` )** intentional in the second snippet?
![Screenshot from 2020-06-29 22-53-41](https://user-images.githubusercontent.com/28972442/86037043-6d326c00-ba5c-11ea-851e-4716c54725a9.png)

Feel free to close the issue if this is expected.

omus:
Thanks for the issue. At the moment `isbounded` and `isunbounded` aren't exported as we will be changing `isopen` to no longer be exported (see https://github.com/invenia/Intervals.jl/pull/100)

The extra backtick is incorrect and should be corrected

LakshyaKhatri:
let me know if you need help with the core library. I'm open to contributions.

PR #126, opened 2020-07-01
Fix span docstring

omus:
Fixed: https://github.com/invenia/Intervals.jl/issues/125

### DIFF
--- a/src/docstrings.jl
+++ b/src/docstrings.jl
@@ -67,7 +67,7 @@ non-negative value while intervals with any unbounded endpoints will throw an
 
 To avoid having to capture the exception use the pattern:
 ```julia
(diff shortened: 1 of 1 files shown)


omus:
Didn't quite work. Interpolation doesn't take into account scope of `Main`

src/docstrings.jl
```diff
@@ -67,7 +67,7 @@ non-negative value while intervals with any unbounded endpoints will throw an
 
 To avoid having to capture the exception use the pattern:
 ```julia
-!isbounded(interval) ? span(interval) : infinity`
+Intervals.isbounded(interval) ? span(interval) : infinity
```
omus, line 70: Removed negation on purpose. Was a flaw in the original docs

PR #128, opened 2020-07-10
Add parsing support for Interval

### LINKED ISSUE #116: Support parsing intervals from strings
It would be very nice to have `parse(::Type{Interval{T}}, ::AbstractString)` which could take strings of the form `"[1,2)"`. I expect we'll need to provide the element type so that we can determine how we should parse the endpoint values.

One tricky part of this will be to determine where the delimiter is as `,` or `..` could be used in the endpoint value. Possibly in scenarios where it is ambiguous we should require quotes around the value.

omus:
Adds parsing support for `Interval{T}`. I ended up just using comma as the parsing delimiter here as I expect it's the syntax most people would expect. Because this PR currently only supports the comma delimiter I changed the printing of intervals to also use a comma. Using a comma for the delimiter when printing has a chance of having a user think that a printed closed-interval is actually a Vector (e.g. `[1, 2]`). We could work around this by leaving printing using ` .. ` and just support both delimiters when parsing.

The parsing performance seems pretty good here despite using a regex. I expect we can improve the parsing performance further but I'm happy with this for now:
```julia
julia> using Intervals, BenchmarkTools

julia> @btime parse(Interval{Int}, "[1,2)")
  731.831 ns (13 allocations: 656 bytes)
Interval{Int64,Closed,Open}(1, 2)

julia> @btime parse(Int, "1")
  26.491 ns (0 allocations: 0 bytes)
1
```

Fixes #116.

Update: Both parsing of delimiters `..` and `,` are supported now and printing has been left unchanged.


omus:
I know that LibPQ rolled its [own interval parsing code](https://github.com/invenia/LibPQ.jl/blob/c234c6ccfebbabcbbd332b3fb7ddaa965226c832/src/parsing.jl#L395) so I took a look at it to ensure this PR works nicely there. The only issue I see is that LibPQ wants to call it's own parsing function. I'll try to support that.

src/parse.jl
```diff
@@ -0,0 +1,33 @@
+const INTERVAL_REGEX = r"""
+    (\[|\()                # Left bound type (1)
+    (?|"([^"]*)"|([^,]*))  # Left bound value (2)
+    ,\ ?
+    (?|"([^"]*)"|([^,]*))  # Right bound value (3)
+    (\]|\))                # Right bound type (4)
+    """x
+
+function Base.parse(::Type{Interval{T}}, str::AbstractString; element_parser=parse) where T
```
omus, line 9: I thought about having the keyword called `el_parser` but that reads as bad Spanish to me

mattBrzezinski:
> The parsing performance seems pretty good here despite using a regex.

Are regexes in Julia inherently slow?

mattBrzezinski:
Just needs a version bump, I assume we would want to bring along a few of the other CRs across the line before making a new release though.

omus:
> > The parsing performance seems pretty good here despite using a regex.
> 
> Are regexes in Julia inherently slow?

The issue isn't that regexes in Julia are slow but rather that regexes in general are slower than if you write a custom lower level parser. This is mostly due to the flexibility of regex and backtracking.

omus:
> Just needs a version bump, I assume we would want to bring along a few of the other CRs across the line before making a new release though.

For GitHub packages I've been bumping the version number outside of the PRs which avoids confusion on whether it has already been bumped. This will be 1.4.0 and I'll probably tag it soon after.

mattBrzezinski:
> > Just needs a version bump, I assume we would want to bring along a few of the other CRs across the line before making a new release though.
> 
> For GitHub packages I've been bumping the version number outside of the PRs which avoids confusion on whether it has already been bumped. This will be 1.4.0 and I'll probably tag it soon after.

Yeah that's what I meant, @iamed2 converted me into this practice :P

omus:
Quick aside. I benchmarked using [possessive quantifiers](https://perldoc.perl.org/perlretut.html#Possessive-quantifiers) to prevent backtracking and there was no performance improvement in this case

src/parse.jl
```diff
@@ -0,0 +1,33 @@
+const INTERVAL_REGEX = r"""
+    (\[|\()                # Left bound type (1)
```
mjram0s, line 3: These comments are very helpful :)

omus: Do the numbers make sense too? You can make named groups with regex but that doesn't work with alternative captures (`?|`)

mjram0s: Yes!

iamed2:
IMO this should support ".." and we should continue to print "..". This is the standard notation for intervals in Julia.

omus:
> IMO this should support ".." and we should continue to print "..". This is the standard notation for intervals in Julia.

I'll make this change. I was starting to worry about how this could break tests which check printing.

omus:
Parsing now supports both `..` and `,` delimiters. I also improved quoting support by allowing an escaped quote character

omus:
There's room for performance improvements here. I'll make a separate issue about that to be done later and get this feature out now

test/interval.jl
```diff
@@ -646,4 +646,100 @@ isinf(::TimeType) = false
...
+        @testset "libpq" begin
+            parse(Interval{Int}, "[\"1\",\"\")") == Interval(1, nothing)
+        end
+
+        @testset "test values" begin
+            for (left, right, _) in test_values, (lb, rb) in product(('[', '('), (']', ')'))
+                T = promote_type(typeof(left), typeof(right))
+
+                # Skip certain types that cannot be parsed at this time
+                T == Char && continue
```
omus, line 735: Not critical but with the introduction of `element_parser` we can avoid skipping this test

omus:
Issue about improving performance: https://github.com/invenia/Intervals.jl/issues/131. The last round of regex changes didn't seem to negatively impact performance:
```julia
julia> @btime parse(Interval{Int}, "[1,2)")
  734.102 ns (13 allocations: 656 bytes)
Interval{Int64,Closed,Open}(1, 2)
```

PR #129, opened 2020-07-10
Support deserialize on previous interval structs

omus:
Added to allow JLSO files saved with the older intervals structure to load transparently. Note that Julia serialization isn't guaranteed to work across versions of Julia but this should work on the same version of Julia with different versions of Intervals.

### DIFF
--- a/test/anchoredinterval.jl
+++ b/test/anchoredinterval.jl
@@ -735,4 +735,19 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
         ai = AnchoredInterval{Day(1)}(zdt)
         @test timezone(ai) == tz"America/Winnipeg"
     end
+
+    @testset "legacy deserialization" begin
+        # Serialized string generated on Intervals@1.2 with:
+        # `julia --project -E 'using Serialization, Intervals; sprint(serialize, AnchoredInterval{-1,Int}(2, true, false))'`.
+        buffer = IOBuffer(
+            SERIALIZED_HEADER *
+            "\x004\x10\x01\x10AnchoredInterval\x1f\v՞\x84\xec\xf7-`\x87\xbb" *
+            "S\xe1Á\x88A\xd8\x01\tIntervalsD\x02\0\0\x001\xff\xff\xff\xff\0\b\xe14\x10" *
+            "\x01\vInclusivity\x1f\v՞\x84\xec\xf7-`\x87\xbbS\xe1Á\x88A\xd8,\x02\0DML"
+        )
+
+        interval = deserialize(buffer)
+        @test interval isa AnchoredInterval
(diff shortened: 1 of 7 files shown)


test/anchoredinterval.jl
```diff
@@ -735,4 +735,18 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
         ai = AnchoredInterval{Day(1)}(zdt)
         @test timezone(ai) == tz"America/Winnipeg"
     end
+
+    @testset "legacy deserialization" begin
+        # Generated with `sprint(deserialize, AnchoredInterval{-1,Int}(2, true, false)` on
```
omus, line 740:
```suggestion
        # Generated with `sprint(deserialize, AnchoredInterval{-1,Int}(2, true, false))` on
```

test/interval.jl
```diff
@@ -631,4 +631,17 @@ isinf(::TimeType) = false
         ]
         @test union(intervals) == [Interval{Open, Closed}(-100, -1)]
     end
+
+    @testset "legacy deserialization" begin
+        # Generated with `sprint(deserialize, Interval(1, 2, true, false)` on Intervals@1.2
```
omus, line 636:
```suggestion
        # Generated with `sprint(deserialize, Interval(1, 2, true, false))` on Intervals@1.2
```

omus:
It occurred to me after I pushed this I should check every Julia minor release that this works. I expect I may need to have different test data per VERSION.

Also, it appear on 1.4 I'll need to use `Inclusivity` as the field type.

omus:
I noticed some deprecation warnings on Julia 1.4 while working on this change. Suppressing the deprecation warnings using a keyword in the case had a strange error about keywords not being supported for the function. I'll investigate further as this is very strange behaviour

omus:
> I noticed some deprecation warnings on Julia 1.4 while working on this change. Suppressing the deprecation warnings using a keyword in the case had a strange error about keywords not being supported for the function. I'll investigate further as this is very strange behaviour

I forgot that `Inclusivity` in the tests is [setup to always test for deprecations](https://github.com/invenia/Intervals.jl/blob/master/test/inclusivity.jl#L1). As that function doesn't accept keywords that is why I was seeing an error. As for why I saw any deprecation warnings at all: that was due to me commenting out the "inclusivity.jl" tests to speedup testing while I was debugging my serialization code. Lessons learned.

omus:
As the CI doesn't test all the versions of Julia I have conditions for I'm manually testing each minor version of Julia from 1.0 to validate the custom deserialization tests are working correctly.

omus:
These changes work confirmed to work with Julia versions: 1.0.5, 1.1.1, 1.2.0, 1.3.1, 1.4.2, and 1.5.0-rc1

PR #138, opened 2020-08-14
add inclusivity default info to the docstrings for HourEnding and HourBeginning

### LINKED ISSUE #77: Add to docstrings HourEnding and HourBegining have default inclusivity
To me it seems like a natural assumption that
`HourEnding` would have inclusivity of `false, true`,
and `HourBeginning` would have inclusivity of `true, false`

but I am not sure if I am typical or not.

mzgubic:
Closes #77

### DIFF
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -118,7 +118,10 @@ AnchoredInterval{P}(anchor::T) where {P,T} = AnchoredInterval{P,T}(anchor)
     HourEnding{T<:TimeType, L, R} <: AbstractInterval{T}
 
 A type alias for `AnchoredInterval{Hour(-1), T}` which is used to denote a 1-hour period of
-time which ends at a time instant (of type `T`). Inclusivity defaults to {Open, Closed}.
+time which ends at a time instant (of type `T`).
+
+When constructing an instance of `HourEnding{T}` the resulting interval will right-closed
+(of type `HourEnding{T,Open,Closed}`).
 """
 const HourEnding{T,L,R} = AnchoredInterval{Hour(-1), T, L, R} where {T, L <: Bounded, R <: Bounded}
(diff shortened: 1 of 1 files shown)


src/anchoredinterval.jl
```diff
@@ -118,7 +118,7 @@ AnchoredInterval{P}(anchor::T) where {P,T} = AnchoredInterval{P,T}(anchor)
     HourEnding{T<:TimeType, L, R} <: AbstractInterval{T}
 
 A type alias for `AnchoredInterval{Hour(-1), T}` which is used to denote a 1-hour period of
-time which ends at a time instant (of type `T`).
+time which ends at a time instant (of type `T`). Inclusivity defaults to {Open, Closed}.
```
omus, line 121:
```suggestion
time which ends at a time instant (of type `T`).

When constructing an instance of `HourEnding{T}` the resulting interval will right-closed
(of type `HourEnding{T,Open,Closed}`).
```

src/anchoredinterval.jl
```diff
@@ -129,7 +129,7 @@ HourEnding(anchor::T) where T = HourEnding{T}(anchor)
     HourBeginning{T<:TimeType, L, R} <: AbstractInterval{T}
 
 A type alias for `AnchoredInterval{Hour(1), T}` which is used to denote a 1-hour period of
-time which begins at a time instant (of type `T`).
+time which begins at a time instant (of type `T`). Inclusivity defaults to {Closed, Open}.
```
omus, line 132:
```suggestion
time which begins at a time instant (of type `T`).

When constructing an instance of `HourBeginning{T}` the resulting interval will left-closed
(of type `HourBeginning{T,Closed,Open}`).
```

mzgubic:
> I like the addition but it needed some additional clarification. Let me know if these make sense to you

Thanks for the review, yes it is much clearer this way.

PR #141, opened 2020-09-14
Create min/max functions that take bounds into account

### LINKED ISSUE #137: Create min/max functions that take bounds into account
It would be nice to be able to get the first and last item within the bounds an Interval to some precision.
Ex.
```
julia> function min(i::Interval{T, Open}; precision=eps(T)) where T
       return first(i) + precision
end 

julia> function min(i::Interval{T, Closed}; precision=eps(T)) where T
       return first(i)
end

julia> min(Interval{Open, Open}(1.0, 2.0); precision=0.1)
1.1

julia> min(Interval{Closed, Closed}(1.0, 2.0); precision=0.1)
1.0
```

Arvind-Maan:
fixes #137

### DIFF
--- a/src/docstrings.jl
+++ b/src/docstrings.jl
@@ -107,29 +107,27 @@ unbounded.
 isbounded(::AbstractInterval)
 
 """
-    minimum(interval::AbstractInterval{T}; increment::Any) -> T
+    minimum(interval::AbstractInterval{T}; [increment]) -> T
 
-The minimum value in the interval.
+The minimum value in the interval contained within the interval.
 
-If left bound is closed, returns `first(interval)`.
-If left bound is unbounded, returns `typemin(T)`
+If left-closed, returns `first(interval)`.
+If left-unbounded, returns minimum value possible for type `T`.
 
-On an empty interval or when the increment results in `first(interval) + increment ∉ interval`, returns BoundsError.
-
-When T <: AbstractFloat and no increment is passed, defaults to returning `nextfloat(first(interval))`
+A `BoundsError` is thrown for empty intervals or when the increment results in a minimum value
+not-contained by the interval.
 """
-minimum(::AbstractInterval; increment::Any)
+minimum(::AbstractInterval; increment)
 
 """
     maximum(interval::AbstractInterval{T}; increment::Any) -> T
 
-The maximum value in the interval.
+The maximum value in the interval contained within the interval.
 
-If right bound is closed, returns `last(interval)`.
-If right bound is unbounded, returns `typemax(T)`
+If right-closed, returns `last(interval)`.
+If right-unbounded, returns maximum value possible for type `T`.
 
-On an empty interval or when the increment results in `last(interval) - increment ∉ interval`, returns BoundsError.
-
-When T <: AbstractFloat and no increment is passed, defaults to returning `prevfloat(last(interval))`
+A `BoundsError` is thrown for empty intervals or when the increment results in a maximum value
+not-contained by the interval.
 """
-maximum(::AbstractInterval; increment::Any)
+maximum(::AbstractInterval; increment)
--- a/src/docstrings.jl
+++ b/src/docstrings.jl
@@ -109,9 +109,10 @@ isbounded(::AbstractInterval)
 """
     minimum(interval::AbstractInterval{T}; [increment]) -> T
 
-The minimum value in the interval contained within the interval.
+The minimum value in the interval contained within the `interval`.
 
 If left-closed, returns `first(interval)`.
+If left-open, returns `first(interval) + eps(first(interval))`
 If left-unbounded, returns minimum value possible for type `T`.
 
 A `BoundsError` is thrown for empty intervals or when the increment results in a minimum value
@@ -122,9 +123,10 @@ minimum(::AbstractInterval; increment)
 """
     maximum(interval::AbstractInterval{T}; increment::Any) -> T
 
-The maximum value in the interval contained within the interval.
+The maximum value in the interval contained within the `interval`.
 
 If right-closed, returns `last(interval)`.
+If right-open, returns `first(interval) + eps(first(interval))`
 If right-unbounded, returns maximum value possible for type `T`.
 
 A `BoundsError` is thrown for empty intervals or when the increment results in a maximum value
--- a/src/docstrings.jl
+++ b/src/docstrings.jl
@@ -105,3 +105,31 @@ Note using `!isbounded` is commonly used to determine if any end of the interval
 unbounded.
 """
 isbounded(::AbstractInterval)
+
+"""
+    minimum(interval::AbstractInterval{T}; increment::Any) -> T
+
+The minimum value in the interval.
+
+If left bound is closed, returns `first(interval)`.
+If left bound is unbounded, returns `typemin(T)`
+
+On an empty interval or when the increment results in `first(interval) + increment ∉ interval`, returns BoundsError.
+
+When T <: AbstractFloat and no increment is passed, defaults to returning `nextfloat(first(interval))`
+"""
+minimum(::AbstractInterval; increment::Any)
+
+"""
+    maximum(interval::AbstractInterval{T}; increment::Any) -> T
+
+The maximum value in the interval.
+
+If right bound is closed, returns `last(interval)`.
+If right bound is unbounded, returns `typemax(T)`
+
+On an empty interval or when the increment results in `last(interval) - increment ∉ interval`, returns BoundsError.
+
+When T <: AbstractFloat and no increment is passed, defaults to returning `prevfloat(last(interval))`
+"""
+maximum(::AbstractInterval; increment::Any)
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -189,10 +189,12 @@ isbounded(interval::AbstractInterval{T,L,R}) where {T,L,R} = L !== Unbounded && 
 
 # Minimum and maximum calls for Closed and Unbounded Intervals, just return first and last.
 function Base.minimum(interval::AbstractInterval{T,L,R}; increment=nothing) where {T,L,R}
+    first(interval) === nothing && return typemin(T)
     return first(interval)
 end
 
 function Base.maximum(interval::AbstractInterval{T,L,R}; increment=nothing) where {T,L,R}
+    last(interval) === nothing && return typemax(T)
     return last(interval)
 end
 
@@ -208,12 +210,12 @@ end
 
 # Minimum and maximum calls for open bounds and Integer types, return open side +- increment
 # Default increment val for Integers is one(T)
-function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=one(T)) where {T<:Integer,R}
-    return first(interval) + increment
+function Base.minimum(interval::AbstractInterval{T,Open,R}) where {T<:Integer,R}
+    return minimum(interval, increment=one(T))
 end
 
-function Base.maximum(interval::AbstractInterval{T,L,Open}; increment=one(T)) where {T<:Integer,L}
-    return last(interval) - increment
+function Base.maximum(interval::AbstractInterval{T,L,Open}) where {T<:Integer,L}
+    return maximum(interval, increment=one(T))
 end
 
 # Minimum and maximum calls for open bounds and Float types, return open side +- increment
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -171,6 +171,22 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
     return R !== Unbounded ? interval.last : nothing
 end
 
+function Base.min(interval::Interval{T,L,R}; precision) where {T,L,R}
+    return Base.first(interval)
+end
+
+function Base.min(interval::Interval{T,Open,R}; precision) where {T,R}
+    return interval.first + precision
+end
+
+function Base.max(interval::Interval{T,L,R}; precision) where {T,L,R}
+    return Base.last(interval)
+end
+
+function Base.max(interval::Interval{T,L,Open}; precision) where {T,L}
+    return interval.last - precision
+end
+
 function span(interval::Interval)
     if isbounded(interval)
         interval.last - interval.first
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -193,11 +193,11 @@ end
 
 function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=eps(T)) where {T,R}
     isempty(interval) && throw(BoundsError(interval, 0))
-    min_val = first(interval)
+    min_val = first(interval) + increment
     # Since intervals can't have NaN, we can just use !isfinite to check if infinite
     !isfinite(min_val) && return typemin(T)
-    min_val + increment ∈ interval && return min_val + increment
-    throw(BoundsError(interval, min_val + increment))
+    min_val ∈ interval && return min_val
+    throw(BoundsError(interval, min_val))
 end
 
 function Base.minimum(interval::AbstractInterval{T,Open,R}) where {T<:Integer,R}
@@ -223,11 +223,11 @@ end
 
 function Base.maximum(interval::AbstractInterval{T,L,Open}; increment=eps(T)) where {T,L}
     isempty(interval) && throw(BoundsError(interval, 0))
-    max_val = last(interval)
+    max_val = last(interval) - increment
     # Since intervals can't have NaN, we can just use !isfinite to check if infinite
     !isfinite(max_val) && return typemax(T)
-    max_val - increment ∈ interval && return max_val - increment
-    throw(BoundsError(interval, max_val - increment))
+    max_val ∈ interval && return max_val
+    throw(BoundsError(interval, max_val))
 end
 
 function Base.maximum(interval::AbstractInterval{T,L,Open}) where {T<:Integer,L}
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -201,11 +201,19 @@ end
 # Minimum and maximum calls for open bounds and ambigous types, return open side +- increment
 # Default increment val is eps(T)
 function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=eps(T)) where {T,R}
-    return first(interval) + increment
+    isempty(interval) && return nothing
+    min_val = first(interval)
+    !isfinite(min_val) && return typemin(T)
+    min_val + increment ∈ interval && return min_val + increment
+    throw(DomainError(interval, "The increment is too large, $(min_val + increment) is not in the interval"))
 end
 
 function Base.maximum(interval::AbstractInterval{T,L,Open}; increment=eps(T)) where {T,L}
-    return last(interval) - increment
+    isempty(interval) && return nothing
+    max_val = last(interval)
+    !isfinite(max_val) && return typemax(T)
+    max_val - increment ∈ interval && return max_val - increment
+    throw(DomainError(interval, "The increment is too large, $(max_val - increment) is not in the interval"))
 end
 
 # Minimum and maximum calls for open bounds and Integer types, return open side +- increment
@@ -221,13 +229,28 @@ end
 # Minimum and maximum calls for open bounds and Float types, return open side +- increment
 # Default increment val for Floats is nothing, which returns nextfloat or prevfloat (i.e., +- eps())
 function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=nothing) where {T<:AbstractFloat,R}
+    isempty(interval) && return nothing
     min_val = first(interval)
-    return increment !== nothing ? min_val + increment : nextfloat(min_val)
+    # Since intervals can't have NaN, we can just use !isfinite to check if infinite
+    next_val = if !isfinite(min_val) || increment === nothing || increment == eps(T)
+        nextfloat(min_val)
+    else
+        min_val + increment
+    end
+    next_val ∈ interval && return next_val
+    throw(DomainError(interval, "The increment is too large, $next_val is not in the interval"))
 end
 
 function Base.maximum(interval::AbstractInterval{T,L,Open}; increment=nothing) where {T<:AbstractFloat,L}
+    isempty(interval) && return nothing
     max_val = last(interval)
-    return increment !== nothing ? max_val - increment : prevfloat(max_val)
+    next_val = if !isfinite(max_val) || increment === nothing || increment == eps(T)
+        prevfloat(max_val)
+    else
+        max_val - increment
+    end
+    next_val ∈ interval && return next_val
+    throw(DomainError(interval, "The increment is too large, $next_val is not in the interval"))
 end
 
 ##### CONVERSION #####
@@ -243,6 +266,7 @@ function Base.convert(::Type{T}, interval::Interval{T}) where T
 end
 
 ##### DISPLAY #####
+
 
 function Base.show(io::IO, interval::Interval{T,L,R}) where {T,L,R}
     if get(io, :compact, false)
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -179,6 +179,10 @@ function Base.minimum(interval::Interval{T,Open,R}; increment=eps(T)) where {T,R
     return first(interval) + increment
 end
 
+function Base.minimum(interval::Interval{T,L,R}; precision=one(T)) where {T <: Integer, L, R}
+    return minimum(interval; precision=precision)
+end
+
 function Base.maximum(interval::Interval{T,L,Closed}; increment=nothing) where {T,L}
     return last(interval)
 end
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -171,6 +171,22 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
     return R !== Unbounded ? interval.last : nothing
 end
 
+function span(interval::Interval)
+    if isbounded(interval)
+        interval.last - interval.first
+    else
+        throw(DomainError(
+            "unbounded endpoint(s)",
+            "Unable to determine the span of an non-bounded interval",
+        ))
+    end
+end
+
+isclosed(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Closed && R === Closed
+Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R === Open
+isunbounded(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Unbounded && R === Unbounded
+isbounded(interval::AbstractInterval{T,L,R}) where {T,L,R} = L !== Unbounded && R !== Unbounded
+
 function Base.minimum(interval::Interval{T,Closed,R}; increment=nothing) where {T,R}
     return first(interval)
 end
@@ -204,22 +220,6 @@ function Base.maximum(interval::Interval{T,L,Open}; increment=nothing) where {T<
     max_val = last(interval)
     return increment !== nothing ? max_val - increment : prevfloat(max_val)
 end
-
-function span(interval::Interval)
-    if isbounded(interval)
-        interval.last - interval.first
-    else
-        throw(DomainError(
-            "unbounded endpoint(s)",
-            "Unable to determine the span of an non-bounded interval",
-        ))
-    end
-end
-
-isclosed(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Closed && R === Closed
-Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R === Open
-isunbounded(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Unbounded && R === Unbounded
-isbounded(interval::AbstractInterval{T,L,R}) where {T,L,R} = L !== Unbounded && R !== Unbounded
 
 ##### CONVERSION #####
 
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -171,7 +171,7 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
     return R !== Unbounded ? interval.last : nothing
 end
 
-function Base.min(interval::Interval{T,L,R}; precision) where {T,L,R}
+function Base.min(interval::Interval{T,L,R}; precision=nothing) where {T,L,R}
     return Base.first(interval)
 end
 
@@ -179,7 +179,7 @@ function Base.min(interval::Interval{T,Open,R}; precision) where {T,R}
     return interval.first + precision
 end
 
-function Base.max(interval::Interval{T,L,R}; precision) where {T,L,R}
+function Base.max(interval::Interval{T,L,R}; precision=nothing) where {T,L,R}
     return Base.last(interval)
 end
 
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -171,20 +171,20 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
     return R !== Unbounded ? interval.last : nothing
 end
 
-function Base.min(interval::Interval{T,L,R}; precision=nothing) where {T,L,R}
-    return Base.first(interval)
+function Base.minimum(interval::Interval{T,Closed,R}; precision=nothing) where {T,R}
+    return first(interval)
 end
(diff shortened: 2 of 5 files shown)


src/interval.jl
```diff
@@ -171,6 +171,22 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
     return R !== Unbounded ? interval.last : nothing
 end
 
+function Base.min(interval::Interval{T,L,R}; precision) where {T,L,R}
+    return Base.first(interval)
+end
+
+function Base.min(interval::Interval{T,Open,R}; precision) where {T,R}
```
Arvind-Maan, line 178: I deliberately didn't set precision as anything, since I feel like having a default could cause issues with different types. 

For example, [in the issue the example](https://github.com/invenia/Intervals.jl/issues/137#issue-678556586) used `eps(..)` as the default but this would cause errors on intervals that weren't floats. 

By leaving this keyword empty it makes the caller _have_ to define the type and how precise they want the function to be.

morris25 (approved):

src/interval.jl
fchorney: If it's a keyword then it's not required to be called, so what would `max(I)` be? where `I` is some interval? 
I guess it would make sense to expect an error if you don't define the precision, but when the interval is closed you don't need to supply a precision, so it seems somewhat inconsistent in that way? 

That being said I'm not really sure how to solve that issue, as you are correct that different types would require different defaults.

Arvind-Maan: max(I) would cause an error since precision wouldn't be defined. 

One workaround I can think of is to define precision as None by default in the min/max functions that are `Unbounded` or `Closed` since precision doesn't matter in those.

This would make the call pass without changing the interface.

fchorney: I guess thinking about it more that should be fine. It makes sense that the function signatures be the same for the closed and open functions even if the precision does nothing on the closed ones. If we wanted viable defaults for precision we'd probably need to define the function for a bunch of different types which may not be worth it at this point.

fchorney (approved):

src/interval.jl
```diff
@@ -171,6 +171,22 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
...
+    return interval.first + precision
+end
+
+function Base.max(interval::Interval{T,L,R}; precision=nothing) where {T,L,R}
+    return Base.last(interval)
+end
+
+function Base.max(interval::Interval{T,L,Open}; precision) where {T,L}
+    return interval.last - precision
```
omus, line 187:
```suggestion
    return last(interval) - precision
```

src/interval.jl
omus: I would say that the default should be `eps` and for types where that value isn't defined it just needs to be passed in explicitly.

src/interval.jl
```diff
@@ -171,6 +171,22 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
     return R !== Unbounded ? interval.last : nothing
 end
 
+function Base.min(interval::Interval{T,L,R}; precision=nothing) where {T,L,R}
```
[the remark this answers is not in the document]
Arvind-Maan: > Since the other `min` function doesn't set a default precision this function shouldn't either.

Since I added the default of eps(T) to the other function, I assume that it's fine to leave this default?

src/interval.jl
```diff
@@ -171,6 +171,22 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
     return R !== Unbounded ? interval.last : nothing
 end
 
+function Base.minimum(interval::Interval{T,Closed,R}; precision=nothing) where {T,R}
```
omus, line 174: Just for consistency:
```suggestion
function Base.minimum(interval::Interval{T,Closed,R}; precision=eps(T)) where {T,R}
```

src/interval.jl
```diff
@@ -171,6 +171,22 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
...
+    return first(interval)
+end
+
+function Base.minimum(interval::Interval{T,Open,R}; precision=eps(T)) where {T,R}
+    return first(interval) + precision
+end
+
+function Base.maximum(interval::Interval{T,L,Closed}; precision=nothing) where {T,L}
```
omus, line 182:
```suggestion
function Base.maximum(interval::Interval{T,L,Closed}; precision=eps(T)) where {T,L}
```

src/interval.jl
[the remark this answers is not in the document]
omus: I'd keep the two methods consistent

src/interval.jl
```diff
@@ -171,6 +171,22 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
...
 end
 
+function Base.minimum(interval::Interval{T,Closed,R}; precision=nothing) where {T,R}
+    return first(interval)
+end
+
+function Base.minimum(interval::Interval{T,Open,R}; precision=eps(T)) where {T,R}
+    return first(interval) + precision
+end
+
```
omus, line 181: You can make things nicer for integers by doing:
```suggestion

function Base.minimum(interval::Interval{T,L,R}; precision=one(T)) where {T <: Integer, L, R}
    return minimum(interval; precision=precision)
end

```

omus: We should also probably special case `AbstractFloat` as with very large floating-point values adding `eps` is inaccurate:
```julia
julia> x = nextfloat(typemin(Float64))
-1.7976931348623157e308

julia> x + eps(Float64)
-1.7976931348623157e308

julia> x + eps(Float64) == x
true

julia> nextfloat(x)
-1.7976931348623155e308

julia> nextfloat(x) == x
false
```

omus:
Another thing I'm thinking about is the name "precision". Currently Julia already has a `precision` function which returns an integer. I think the term `increment` may be better suited

src/interval.jl
omus: Actually, I take this back. You can leave it as `nothing`. I thought Julia was smart enough to optimize this away. That isn't the case.
```julia
julia> f(x=nothing) = 1
f (generic function with 2 methods)

julia> g(x=eps(Int)) = 1
g (generic function with 2 methods)

julia> @code_native f()
	.section	__TEXT,__text,regular,pure_instructions
; ┌ @ REPL[5]:1 within `f'
	movl	$1, %eax
	retq
	nopw	%cs:(%rax,%rax)
; └

julia> @code_native g()
	.section	__TEXT,__text,regular,pure_instructions
; ┌ @ REPL[6]:1 within `g'
	pushq	%rax
	movabsq	$jl_system_image_data, %rax
	movq	%rax, (%rsp)
	movabsq	$jl_apply_generic, %rax
	movabsq	$jl_system_image_data, %rdi
	movq	%rsp, %rsi
	movl	$1, %edx
	callq	*%rax
	ud2
	nop
; └

julia> g()
ERROR: MethodError: no method matching eps(::Type{Int64})
Closest candidates are:
  eps(::Dates.Time) at /Users/omus/Development/Julia/1.3/usr/share/julia/stdlib/v1.3/Dates/src/types.jl:387
  eps(::Dates.Date) at /Users/omus/Development/Julia/1.3/usr/share/julia/stdlib/v1.3/Dates/src/types.jl:386
  eps(::Dates.DateTime) at /Users/omus/Development/Julia/1.3/usr/share/julia/stdlib/v1.3/Dates/src/types.jl:385
  ...
Stacktrace:
 [1] g() at ./REPL[6]:1
 [2] top-level scope at REPL[9]:1
```

src/interval.jl
omus: See https://github.com/invenia/Intervals.jl/pull/141/files#r488735402

src/interval.jl
```diff
@@ -171,6 +171,38 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
...
+function Base.minimum(interval::Interval{T,Open,R}; increment=eps(T)) where {T<:AbstractFloat,R}
+    return nextfloat(first(interval))
+end
+
+function Base.maximum(interval::Interval{T,L,Open}; increment=eps(T)) where {T<:AbstractFloat,L}
+    return prevfloat(last(interval))
+end
+
```
omus, line 205: Can we move these definitions to after the `isbounded` function definition? Also, these functions should probably use `AbstractInterval` instead of `Interval` so that this also works for `AnchoredIntervals`.

test/interval.jl
```diff
@@ -125,12 +125,33 @@ isinf(::TimeType) = false
...
+                # the value we compare to min/max depends on if the bound is open/closed
+                # for floats, adding/subtracting eps(Float) is inaccurate for large vals
+                # so we must use next/prev float
+
+                min_comp = if a in interval
+                    first(interval)
+                else
+                    typeof(first(interval)) <: AbstractFloat ?
+                        nextfloat(first(interval)) :
+                        first(interval) + p
```
omus, line 143:
```suggestion
                    val = first(interval)
                    eltype(interval) isa AbstractFloat ? nextfloat(val) : val + p
```

test/interval.jl
```diff
@@ -125,12 +125,33 @@ isinf(::TimeType) = false
...
             for (L, R) in BOUND_PERMUTATIONS
                 interval = Interval{L, R}(a, b)
 
                 @test first(interval) == a
                 @test last(interval) == b
+                # the value we compare to min/max depends on if the bound is open/closed
+                # for floats, adding/subtracting eps(Float) is inaccurate for large vals
+                # so we must use next/prev float
+
+                min_comp = if a in interval
```
omus, line 138:
```suggestion

                # The value we compare to min/max depends on if the bound is open/closed
                # for floats, adding/subtracting eps(Float) is inaccurate for large vals
                # so we must use next/prev float
                min_comp = if a in interval
```

src/interval.jl
omus: We should also have tests for `AnchoredInterval`s and intervals using `Unbounded`

src/interval.jl
```diff
@@ -171,6 +171,38 @@ function Base.last(interval::Interval{T,L,R}) where {T,L,R}
...
+    return last(interval) - increment
+end
+
+function Base.minimum(interval::Interval{T,Open,R}; increment=eps(T)) where {T<:AbstractFloat,R}
+    return nextfloat(first(interval))
+end
+
+function Base.maximum(interval::Interval{T,L,Open}; increment=eps(T)) where {T<:AbstractFloat,L}
+    return prevfloat(last(interval))
+end
```
omus, line 204:
```suggestion
function Base.minimum(interval::Interval{T,Open,R}; increment=nothing) where {T<:AbstractFloat,R}
    min_val = first(interval)
    return increment !== nothing ? min_val + increment : nextfloat(min_val)
end

function Base.maximum(interval::Interval{T,L,Open}; increment=nothing) where {T<:AbstractFloat,L}
    max_val = last(interval)
    return increment !== nothing ? max_val - increment : prevfloat(max_val)
end
```

test/interval.jl
omus: Actually, this extra logic can be removed if `AbstractFloat` respects explicitly set increment values

test/interval.jl
```diff
@@ -125,12 +125,43 @@ isinf(::TimeType) = false
...
+        for (a, b, p) in append!(accessor_specific_vals, test_values)
             for (L, R) in BOUND_PERMUTATIONS
                 interval = Interval{L, R}(a, b)
 
                 @test first(interval) == a
                 @test last(interval) == b
+
+                # If p=eps(Float64) then we treat it as if p=nothing.
+                # This is because adding eps to floats is not accurate and we must use nextfloat and prevfloat.
+                p = p == eps() ? nothing : p
```
omus, line 145: Doesn't this result in `(-10.0, 10.0, nothing)` and `(-10.0, 10.0, eps(Float64))` being identical tests?

Arvind-Maan: Whooops, this is old code that I accidentally pushed. Will resolve.

src/interval.jl
```diff
@@ -187,6 +187,47 @@ Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R =
...
+
+# Minimum and maximum calls for open bounds and ambigous types, return open side +- increment
+# Default increment val is eps(T)
+function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=eps(T)) where {T,R}
+    return first(interval) + increment
+end
+
+function Base.maximum(interval::AbstractInterval{T,L,Open}; increment=eps(T)) where {T,L}
+    return last(interval) - increment
+end
```
Arvind-Maan, line 207: What should I do in the following case:

``` julia
first(interval) + increment > last(interval)
last(interval) - increment < first(interval)
```

should I be returning `nothing` in these cases? Since they are ∉ the interval

test/interval.jl
```diff
@@ -149,6 +177,58 @@ isinf(::TimeType) = false
...
+            (nothing, nothing, 1, typeof(1), Unbounded, Unbounded),
+            (nothing, nothing, 0.01, Float64, Unbounded, Unbounded),
+            (nothing, nothing, 1, Char, Unbounded, Unbounded),
+            (nothing, nothing, Day(1), Day, Unbounded, Unbounded),
+            (nothing, nothing, Millisecond(1), DateTime, Unbounded, Unbounded),
+
+            # test adding eps() with unbounded
+            (-10.0, nothing, nothing, Float64, Open, Unbounded),
+            (nothing, 10.0, eps(Float64), Float64, Unbounded, Open),
+        ]
```
omus, line 199: May as well construct actual intervals in this case as you have all of the required information:
```suggestion
        unbounded_test_values = [
            # one side unbounded with different types
            (Interval{Open,Unbounded}(-10, nothing), 1)
            (Interval{Unbounded,Closed}(nothing, 1.0), 0.01),
            (Interval{Unbounded,Open}(nothing, 'z') 1),
            (Interval{Closed,Unbounded}(Date(2013, 2, 13), nothing), Day(1)),
            (Interval{Open,Unbounded}(DateTime(2016, 8, 11, 0, 30), nothing), Millisecond(1)),

            # both sides unbounded different types
            (Interval{Int}(nothing, nothing), 1),
            (Interval{Float64}(nothing, nothing), 0.01),
            (Interval{Char}(nothing, nothing), 1),
            (Interval{Day}(nothing, nothing), Day(1)),
            (Interval{DateTime}(nothing, nothing), Millisecond(1)),

            # test adding eps() with unbounded
            (Interval{Open,Unbounded}(-10.0, nothing), nothing),
            (Interval{Unbounded,Open}(nothing, 10.0), eps(Float64)),
        ]
```

test/interval.jl
```diff
@@ -124,19 +124,47 @@ isinf(::TimeType) = false
...
+
+                ma = if b ∈ interval
+                    last(interval)
+                else
+                    v = last(interval)
+                    unit === nothing && eltype(interval) <: AbstractFloat ? prevfloat(v) : v - unit
+                end
+
+                @test minimum(interval; increment=unit) == mi
+                @test maximum(interval; increment=unit) == ma
```
omus, line 160: I think I'd combine these test with what you have below and call the testset "minimum/maximum"

test/interval.jl
```diff
@@ -149,6 +177,58 @@ isinf(::TimeType) = false
...
+        for (a, b, unit, T, L, R) in unbounded_test_values
+            interval = Interval{T,L,R}(a, b)
+
+            @test first(interval) == a
+            @test last(interval) == b
+
+            # The value we compare to min/max depends on if the bound is open/closed.
+            # we can determine if a bound is open/closed by checking if the edge in the interval
+            mi = if a === nothing
+                nothing
```
omus, line 209: I was originally thinking when using `minimum` on an left-unbounded interval you would get the minimum possible value in that interval. When taking into account the element type this could return `typemin(eltype(interval))`. There can definitely be some discussion on what we expect in this case

src/interval.jl
omus: A very good question. As the provided increment is supposed to be the smallest span between to values I was thinking an exception would be reasonable. Looking at empty intervals (e.g. `Interval{Open,Open}(0, 0)`) should probably be similar to using `minimum`/`maximum` on an empty collection which raises an exception:
```julia
julia> maximum(Int[])
ERROR: ArgumentError: reducing over an empty collection is not allowed
```

test/anchoredinterval.jl
```diff
@@ -187,6 +191,8 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
         interval = AnchoredInterval{Day(1)}(startpoint)
         @test first(interval) == startpoint
         @test last(interval) == ZonedDateTime(2018, 3, 12, tz"America/Winnipeg")
+        @test minimum(interval) == startpoint
+        @test maximum(interval, increment=Hour(1)) == ZonedDateTime(2018, 3, 11, 23, tz"America/Winnipeg")
```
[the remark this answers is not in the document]
Arvind-Maan: eps(ZonedDateTime) creates an error unfortunately.

src/interval.jl
morris25: I suppose it would depend if the other side is open or closed. You would expect an `Interval{T, Closed, Open}` to have a maximum regardless of increment.

test/interval.jl
```diff
@@ -144,6 +143,164 @@
...
+                l - unit
+            end
+
+            nv ∈ interval && return nv
+
+            # If we get to this point, the min/max functions throw a DomainError
+            # Since we want our tests to be predictable, we will not throw an error in this helper.
+        end
+        @testset "bounded intervals" begin
+            bounded_test_vals = [
```
Arvind-Maan, line 208: Can't use Infinity.jl since `isfinite`does ` ∞ -  ∞` which causes an error. Could make a workaround for this

nickrobinson251:
As a follow-up to this PR, it'd be nice to resolve https://github.com/invenia/Intervals.jl/issues/90 😊

src/interval.jl
```diff
@@ -187,6 +187,72 @@ Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R =
 isunbounded(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Unbounded && R === Unbounded
 isbounded(interval::AbstractInterval{T,L,R}) where {T,L,R} = L !== Unbounded && R !== Unbounded
 
+# Minimum and maximum calls for Closed and Unbounded Intervals, just return first and last.
+function Base.minimum(interval::AbstractInterval{T,L,R}; increment=nothing) where {T,L,R}
+    first(interval) === nothing && return typemin(T)
+    return first(interval)
+end
```
omus, line 194:
```suggestion
function Base.minimum(interval::AbstractInterval{T,L,R}; increment=nothing) where {T,L,R}
    return L === Unbounded ? typemin(T) : interval.first
end
```
OR use multiple dispatch to remove the conditional check:
```suggestion
Base.minimum(interval::AbstractInterval{T,Unbounded,R}; increment=nothing) where {T,R}
    return typemin(T)
end

Base.minimum(interval::AbstractInterval{T,Closed,R}; increment=nothing) where {T,R}
    return interval.first
end
```
Note: I'm using `interval.first` over `first(interval)` here to avoid the unbounded check `first` does

src/interval.jl
```diff
@@ -187,6 +187,72 @@ Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R =
...
+
+function Base.maximum(interval::AbstractInterval{T,L,R}; increment=nothing) where {T,L,R}
+    last(interval) === nothing && return typemax(T)
+    return last(interval)
+end
+
+# Minimum and maximum calls for open bounds and ambigous types, return open side +- increment
+# Default increment val is eps(T)
+function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=eps(T)) where {T,R}
+    isempty(interval) && return nothing
```
omus, line 204: I think this should error

src/interval.jl
```diff
@@ -187,6 +187,72 @@ Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R =
...
+end
+
+function Base.maximum(interval::AbstractInterval{T,L,R}; increment=nothing) where {T,L,R}
+    last(interval) === nothing && return typemax(T)
+    return last(interval)
+end
+
+# Minimum and maximum calls for open bounds and ambigous types, return open side +- increment
+# Default increment val is eps(T)
+function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=eps(T)) where {T,R}
```
omus, line 203: I'd group the `minimum` functions all together

src/interval.jl
```diff
@@ -187,6 +187,72 @@ Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R =
...
+    return maximum(interval, increment=one(T))
+end
+
+# Minimum and maximum calls for open bounds and Float types, return open side +- increment
+# Default increment val for Floats is nothing, which returns nextfloat or prevfloat (i.e., +- eps())
+function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=nothing) where {T<:AbstractFloat,R}
+    isempty(interval) && return nothing
+    min_val = first(interval)
+    # Since intervals can't have NaN, we can just use !isfinite to check if infinite
+    next_val = if !isfinite(min_val) || increment === nothing || increment == eps(T)
```
omus, line 235: I'd only use `nextfloat` if a increment isn't explicitly given. Even if it is `eps(T)`
```suggestion
    next_val = if !isfinite(min_val) || increment === nothing
```

test/anchoredinterval.jl
```diff
@@ -175,6 +177,9 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
 
         @test first(interval) == Date(2016, 8, 11)
         @test last(interval) == Date(2016, 8, 12)
+        # throws domain error
+        @test_throws DomainError minimum(interval, increment=Day(1)) == first(interval) + Day(1)
+        @test_throws DomainError maximum(interval, increment=Day(1)) == last(interval) - Day(1)
```
omus, line 182:
```suggestion
        @test_throws DomainError minimum(interval, increment=Day(1))
        @test_throws DomainError maximum(interval, increment=Day(1))
```

test/anchoredinterval.jl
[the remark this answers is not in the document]
omus: We should be using the smallest value supported by the element type as that's what `increment` is intended to be set to:
```suggestion
        @test maximum(interval, increment=Millisecond(1)) == last(interval) - Millisecond(1)
```

test/anchoredinterval.jl
```diff
@@ -197,24 +204,32 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
         interval = AnchoredInterval{Day(1)}(startpoint)
         @test first(interval) == startpoint
         @test last(interval) == ZonedDateTime(2018, 11, 5, tz"America/Winnipeg")
+        @test minimum(interval) == startpoint
+        @test maximum(interval, increment=Hour(1)) == ZonedDateTime(2018, 11, 4, 23, tz"America/Winnipeg")
```
omus, line 208:
```suggestion
        @test maximum(interval, increment=Millisecond(1)) == last(interval) - Millisecond(1)
```

test/anchoredinterval.jl
```diff
@@ -197,24 +204,32 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
...
         @test span(interval) == Day(1)
 
         endpoint = ZonedDateTime(2020, 3, 9, 2, tz"America/Winnipeg")
         interval = AnchoredInterval{Day(-1)}(endpoint)
         @test_throws NonExistentTimeError first(interval)
+        @test_throws NonExistentTimeError minimum(interval, increment=Hour(1))
         @test last(interval) == endpoint
+        @test maximum(interval) == endpoint
```
omus, line 216: Be consistent with ordering
```suggestion
        @test_throws NonExistentTimeError first(interval)
        @test last(interval) == endpoint
        @test_throws NonExistentTimeError minimum(interval, increment=Millisecond(1))
        @test maximum(interval) == endpoint
```

test/interval.jl
omus: If you rebase your changes this should no longer be an issue:
```julia
julia> isfinite(∞)
false

julia> Intervals.isfinite(∞)
false
```

Arvind-Maan: Might need some more clarification on this.

I rebased in the traditional sense `git rebase master` and I also updated from master, but it still doesn't work for some reason. 🙁 

If I replicate your calls in the REPL, it works fine as well.

Arvind-Maan:
```julia
julia> using Dates,Infinity, Intervals

julia> x = Interval{Open, Open}(Date(2013, 2, 13), ∞)
Interval{InfExtendedTime{Date},Open,Open}(InfExtendedTime{Date}(Date("2013-02-13")), InfExtendedTime{Date}(∞))

julia> minimum(x, increment=Day(1))
InfExtendedTime{Date}(Date("2013-02-14"))

julia> maximum(x, increment=Day(1))
ERROR: ∞-∞ is undefined
Stacktrace:
 [1] + at /Users/arvindmaan/.julia/packages/Infinity/ezs8D/src/infinite/arithmetic.jl:9 [inlined]
 [2] -(::Infinite, ::Infinite) at /Users/arvindmaan/.julia/packages/Infinity/ezs8D/src/infinite/arithmetic.jl:10
 [3] -(::InfExtendedTime{Date}, ::InfExtendedTime{Date}) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Dates/src/arithmetic.jl:9
 [4] isfinite(::InfExtendedTime{Date}) at /Users/arvindmaan/.julia/dev/Intervals/src/isfinite.jl:3
 [5] maximum(::Interval{InfExtendedTime{Date},Open,Open}; increment::Day) at /Users/arvindmaan/.julia/dev/Intervals/src/interval.jl:228
 [6] top-level scope at REPL[7]:1
```

src/docstrings.jl
```diff
@@ -105,3 +105,31 @@ Note using `!isbounded` is commonly used to determine if any end of the interval
...
+
+"""
+    minimum(interval::AbstractInterval{T}; increment::Any) -> T
+
+The minimum value in the interval.
+
+If left bound is closed, returns `first(interval)`.
+If left bound is unbounded, returns `typemin(T)`
+
+On an empty interval or when the increment results in `first(interval) + increment ∉ interval`, returns BoundsError.
```
omus, line 117:
```suggestion
A `BoundsError` is thrown for empty intervals or when the increment results in a minimum value not-contained by the interval.
```

src/docstrings.jl
```diff
@@ -105,3 +105,31 @@ Note using `!isbounded` is commonly used to determine if any end of the interval
...
+
+When T <: AbstractFloat and no increment is passed, defaults to returning `nextfloat(first(interval))`
```
omus, line 119: I think we can leave this implementation detail out.

src/interval.jl
```diff
@@ -187,70 +187,63 @@ Base.isopen(interval::AbstractInterval{T,L,R}) where {T,L,R} = L === Open && R =
...
-
-# Minimum and maximum calls for open bounds and ambigous types, return open side +- increment
-# Default increment val is eps(T)
 function Base.minimum(interval::AbstractInterval{T,Open,R}; increment=eps(T)) where {T,R}
-    isempty(interval) && return nothing
+    isempty(interval) && throw(BoundsError(interval, 0))
     min_val = first(interval)
+    # Since intervals can't have NaN, we can just use !isfinite to check if infinite
     !isfinite(min_val) && return typemin(T)
     min_val + increment ∈ interval && return min_val + increment
```
omus:
```suggestion
    min_val = interval.first + increment
    # Since intervals can't have NaN, we can just use !isfinite to check if infinite
    !isfinite(min_val) && return typemin(T)
    min_val ∈ interval && return min_val
```

test/interval.jl
omus: The trouble is definitely from `InfExtendedTime` specifically here. I'll see if I can come up with a solution for that

omus: Add the following to the end of `src/isfinite.jl` and everything should be working:
```julia
isfinite(x::Union{Type{T}, T}) where T<:TimeType = Base.isfinite(x)
```

omus:
Thanks for sticking with this. Things are shaping up nicely now :)

test/anchoredinterval.jl
```diff
@@ -175,6 +177,9 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
 
         @test first(interval) == Date(2016, 8, 11)
         @test last(interval) == Date(2016, 8, 12)
+        # throws domain error
```
omus, line 180:
```suggestion
```

test/interval.jl
```diff
@@ -144,6 +143,141 @@
...
+                prevfloat(l)
+            else
+                l - unit
+            end
+
+            nv ∈ interval && return nv
+
+            # If we get to this point, the min/max functions throw a DomainError
+            # Since we want our tests to be predictable, we will not throw an error in this helper.
+        end
```
omus, line 206: As theses helpers pretty much just re-implement minimum/maximum there aren't that helpful with testing. I think we can proceed with this but something to keep in mind for future work.

src/docstrings.jl
```diff
@@ -105,3 +105,29 @@ Note using `!isbounded` is commonly used to determine if any end of the interval
 unbounded.
 """
 isbounded(::AbstractInterval)
+
+"""
+    minimum(interval::AbstractInterval{T}; [increment]) -> T
+
+The minimum value in the interval contained within the interval.
```
omus, line 112:
```suggestion
    minimum(interval::AbstractInterval{T}; [increment]) -> T

The minimum value contained within the `interval`.
```

src/docstrings.jl
```diff
@@ -105,3 +105,29 @@ Note using `!isbounded` is commonly used to determine if any end of the interval
...
+
+A `BoundsError` is thrown for empty intervals or when the increment results in a minimum value
+not-contained by the interval.
+"""
+minimum(::AbstractInterval; increment)
+
+"""
+    maximum(interval::AbstractInterval{T}; increment::Any) -> T
+
+The maximum value in the interval contained within the interval.
```
omus, line 125:
```suggestion
    maximum(interval::AbstractInterval{T}; [increment]) -> T

The maximum value contained within the `interval`.
```

src/docstrings.jl
```diff
@@ -105,3 +105,29 @@ Note using `!isbounded` is commonly used to determine if any end of the interval
 unbounded.
 """
 isbounded(::AbstractInterval)
+
+"""
+    minimum(interval::AbstractInterval{T}; [increment]) -> T
+
+The minimum value in the interval contained within the interval.
+
+If left-closed, returns `first(interval)`.
```
omus, line 114:
```suggestion
If left-closed, returns `first(interval)`.
If left-open, returns `first(interval) + eps(first(interval))`
```

src/docstrings.jl
```diff
@@ -105,3 +105,29 @@ Note using `!isbounded` is commonly used to determine if any end of the interval
...
+not-contained by the interval.
+"""
+minimum(::AbstractInterval; increment)
+
+"""
+    maximum(interval::AbstractInterval{T}; increment::Any) -> T
+
+The maximum value in the interval contained within the interval.
+
+If right-closed, returns `last(interval)`.
```
omus, line 127:
```suggestion
If right-closed, returns `last(interval)`.
If right-open, returns `last(interval) - eps(last(interval))`
```

omus (approved):
With these last changes we should be good.

test/interval.jl
Arvind-Maan: I agree, in hindsight it would've been best to separate the tests per edge case and avoid having to do this.

src/docstrings.jl
omus: Looks like this change was missed?

src/docstrings.jl
omus: Looks like this change was missed?

omus:
PR was meant to be squashed. Squashed commit is https://github.com/invenia/Intervals.jl/commit/c37cbb77cab32c842cd482ce3170eeb18b64d9e9

PR #142, opened 2020-09-14
Refactor generic `isfinite` support

omus:
As `isfinite(::Char)` is something that [isn't desired to be supported in Base](https://github.com/JuliaLang/julia/pull/36380#issuecomment-690646495) I'll refactor the Intervals.jl code to make internal support for this more reasonable.

### DIFF
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -24,6 +24,7 @@ Base.eltype(::AbstractInterval{T}) where {T} = T
 Base.broadcastable(x::AbstractInterval) = Ref(x)
 bounds_types(x::AbstractInterval{T,L,R}) where {T,L,R} = (L, R)
 
+include("isfinite.jl")
 include("endpoint.jl")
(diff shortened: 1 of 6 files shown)


src/Intervals.jl
```diff
@@ -24,6 +24,7 @@ Base.eltype(::AbstractInterval{T}) where {T} = T
 Base.broadcastable(x::AbstractInterval) = Ref(x)
 bounds_types(x::AbstractInterval{T,L,R}) where {T,L,R} = (L, R)
 
+include("isfinite.jl")
```
mattBrzezinski, line 27: Would this be worthwhile to maybe export?

mattBrzezinski (approved):

src/Intervals.jl
omus: I definitely don't want to export this as it'll cause conflicts with `Base.isfinite`

PR #143, opened 2020-09-23
Restrict arguments to Number in comparisons between scalars and EndPoint

p-zubieta:
Looking at some packages invalidations, I found some introduced by this one. As far as I can see, much of these methods aren't supposed to work for things other than `Number`s so I think these all should be harmless.

### DIFF
--- a/src/endpoint.jl
+++ b/src/endpoint.jl
@@ -145,8 +145,12 @@ function Base.isless(a::RightEndpoint, b::LeftEndpoint)
 end
 
 # Comparisons between Scalars and Endpoints
-Base.:(==)(a, b::Endpoint) = a == b.endpoint && isclosed(b)
-Base.:(==)(a::Endpoint, b) = b == a
+Base.:(==)(a::T, b::Endpoint{T}) where {T} = a == b.endpoint && isclosed(b)
+Base.:(==)(a::Endpoint{T}, b::T) where {T} = b == a
+Base.:(==)(a::Number, b::Endpoint{<:Number}) = a == b.endpoint && isclosed(b)
+Base.:(==)(a::Endpoint{<:Number}, b::Number) = b == a
+Base.:(==)(a::DatesTypes, b::Endpoint{<:DatesTypes}) = a == b.endpoint && isclosed(b)
+Base.:(==)(a::Endpoint{<:DatesTypes}, b::DatesTypes) = b == a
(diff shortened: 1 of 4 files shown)


pabloferz:
I think I spoke too soon, I see that Intervals are expected, at least from the tests, to work with things like `Dates`. Maybe these really are "irreducible" invalidations, I'll have to think about it a bit more.

src/endpoint.jl
```diff
@@ -145,8 +145,8 @@ function Base.isless(a::RightEndpoint, b::LeftEndpoint)
 end
 
 # Comparisons between Scalars and Endpoints
-Base.:(==)(a, b::Endpoint) = a == b.endpoint && isclosed(b)
-Base.:(==)(a::Endpoint, b) = b == a
+Base.:(==)(a::Number, b::Endpoint) = a == b.endpoint && isclosed(b)
+Base.:(==)(a::Endpoint, b::Number) = b == a
```
omus, line 149: I'm pretty sure we can update this to be this safely (tests will show if that is the case):
```suggestion
Base.:(==)(a::T, b::Endpoint{T}) where T = a == b.endpoint && isclosed(b)
Base.:(==)(a::Endpoint{T}, b::T) where T = b == a
```
I'm not sure these are actually used so we may be able to just drop them.

omus (commented):
Currently the only restriction the element type of `Interval` and `AnchoredInterval` is that the eltype must be comparable with `isless`. We may be able to get away with using promotion to convert a scalar (`x`) into a closed-interval (`Interval(x, x)`) which could work around the method invalidations by not having to define some of these methods. I suspect there will be a performance hit in doing this though.

pabloferz:
From code coverage, it looks like some of the new definitions are not use in the tests and we might get away with removing them. For others, it might be better to add code tests paths.

I'm not sure how does this look for someone more familiar with direct use cases of the package.

omus:
I'm looking into alternative approaches. I think we may be able to fix Base code such the `+` and `-` invalidations no longer occur (see https://github.com/invenia/Intervals.jl/issues/144). As for converting to a scalar I think we can define an alternative function for this behaviour.

pabloferz:
That's great. I guess I will remove the arithmetic code changes. Yes, a custom scalar conversion function sounds reasonable.

Issue #144, opened 2020-09-30
Adding a convert(::Type{T}, ...) where {T} is a source of a large number of invalidations


KristofferC:
The method definition here:

https://github.com/invenia/Intervals.jl/blob/c37cbb77cab32c842cd482ce3170eeb18b64d9e9/src/anchoredinterval.jl#L179-L197

is really bad for latency because it invalidates the `convert(::Type{Any}, x)` method in Base which a huge number of methods depend on. This is a similar issue to https://github.com/JuliaData/CategoricalArrays.jl/issues/177. The invalidations can be seen via:

```jl
julia> using SnoopCompileCore

julia> invalidations = @snoopr using Intervals

julia> using SnoopCompile

julia> trees = invalidation_trees(invalidations);

julia> trees[end]

....
inserting convert(::Type{T}, interval::AnchoredInterval{P, T, L, R} where R<:Intervals.Bounded where L<:Intervals.Bounded) where {P, T} in Intervals at /home/kc/.julia/packages/Intervals/T6hJb/src/anchoredinterval.jl:177 invalidated:
....
                 358: signature Tuple{typeof(convert), Type{Symbol}, Any} triggered MethodInstance for Base.ImmutableDict{Symbol, Any}(::Base.ImmutableDict{Symbol, Any}, ::Any, ::Any) (101 children)
                 359: signature Tuple{typeof(convert), Type{Symbol}, Any} triggered MethodInstance for merge(::NamedTuple{(), Tuple{}}, ::Base.Iterators.Pairs) (394 children)
                 360: signature Tuple{typeof(convert), Type{Union{Dict{String, Any}, Base.TOML.ParserError}}, Any} triggered MethodInstance for recurse_dict!(::Base.TOML.Parser, ::Dict{String, Any}, ::SubArray{String, 1, Vector{String}, Tuple{UnitRange{Int64}}, true}, ::Bool) (536 children)
   backedges: 1: superseding convert(::Type{Any}, x) in Base at essentials.jl:204 with MethodInstance for convert(::Type{Any}, ::Any) (1242 children)
```

These two are also pretty bad:

https://github.com/invenia/Intervals.jl/blob/c37cbb77cab32c842cd482ce3170eeb18b64d9e9/src/endpoint.jl#L148-L149

omus:
An attempt to address this: https://github.com/invenia/Intervals.jl/pull/143

KristofferC:
Thanks, hadn't seen that one.

[...]

PR #145, opened 2020-10-07
Deprecate convert for closed intervals in favor of only

p-zubieta:
Another take on https://github.com/invenia/Intervals.jl/issues/144 as suggested in https://github.com/invenia/Intervals.jl/pull/143. I don't know if this is the best name for this function nor how the presence of the `deprecation` affects removing the method. But throwing the idea here anyway to keep the ball rolling.

### DIFF
--- a/src/anchoredinterval.jl
+++ b/src/anchoredinterval.jl
@@ -176,23 +176,29 @@ span(interval::AnchoredInterval{P}) where P = abs(P)
 
 ##### CONVERSION #####
 
-# Allows an interval to be converted to a scalar when the set contained by the interval only
-# contains a single element.
-function Base.convert(::Type{T}, interval::AnchoredInterval{P,T}) where {P,T}
+function Base.only(interval::AnchoredInterval{P}) where {P}
     if isclosed(interval) && (sign(P) == 0 || first(interval) == last(interval))
         return first(interval)
     else
-        # Remove deprecation in version 2.0.0
-        depwarn(
-            "`convert(T, interval::AnchoredInterval{P,T})` is deprecated for " *
-            "intervals which are not closed with coinciding endpoints. " *
-            "Use `anchor(interval)` instead.",
-            :convert,
-        )
-        return anchor(interval)
+        throw(DomainError(interval,
+            "The interval is not closed with coinciding endpoints, " *
+            "did you meant to use `anchor(interval)`?"
+        ))
+    end
+end
 
-        # TODO: For when deprecation is removed
-        # throw(DomainError(interval, "The interval is not closed with coinciding endpoints"))
+# Remove in version 2.0.0
+function Base.convert(::Type{T}, interval::AnchoredInterval{P,T}) where {P,T}
+    depwarn(
+        "`convert(::Type{T}, interval::AnchoredInterval{P, T})` is deprecated, " *
+        "use `only(interval::AnchoredInterval{P,T})` for closed intervals with " *
+        "coinciding endpoints or `anchor(interval)` otherwise.",
+        :convert,
+    )
+    if isclosed(interval) && (sign(P) == 0 || first(interval) == last(interval))
+        return first(interval)
+    else
+        return anchor(interval)
     end
 end
 
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -237,9 +237,9 @@ end
     only(interval::AbstractInterval)
 
 Returns the only element of a closed interval with coinciding endpoints (throws a
-DomainError otherwise).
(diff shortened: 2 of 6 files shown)


pabloferz:
@omus was something like this what you had in mind [here](https://github.com/invenia/Intervals.jl/pull/143#issuecomment-702372172)?

nalimilan:
Maybe making this a method of `only` would make sense?

KristofferC:
Bump, I'm seeing thousands of invalidations less with this PR (for a big package I am looking at).

pabloferz:
For the latter there is #143.

omus:
> @omus was something like this what you had in mind [here](https://github.com/invenia/Intervals.jl/pull/143#issuecomment-702372172)?

Yes, this is. I was thinking of using an alternative function name like `scalar` but `only` seems quite reasonable

src/anchoredinterval.jl
```diff
@@ -176,23 +176,11 @@ span(interval::AnchoredInterval{P}) where P = abs(P)
 
 ##### CONVERSION #####
 
-# Allows an interval to be converted to a scalar when the set contained by the interval only
-# contains a single element.
-function Base.convert(::Type{T}, interval::AnchoredInterval{P,T}) where {P,T}
+function Base.only(interval::AnchoredInterval{P}) where {P}
```
omus, line 179: Need to properly deprecate the use of `convert`. We can turn around and drop this and other depreciations in version 2

src/anchoredinterval.jl
```diff
@@ -176,23 +176,11 @@ span(interval::AnchoredInterval{P}) where P = abs(P)
...
-            "`convert(T, interval::AnchoredInterval{P,T})` is deprecated for " *
-            "intervals which are not closed with coinciding endpoints. " *
-            "Use `anchor(interval)` instead.",
-            :convert,
-        )
-        return anchor(interval)
-
-        # TODO: For when deprecation is removed
-        # throw(DomainError(interval, "The interval is not closed with coinciding endpoints"))
+        throw(DomainError(interval, "The interval is not closed with coinciding endpoints"))
```
omus, line 183: This change should be reverted with an appropriate change to the depwarn

src/interval.jl
```diff
@@ -248,9 +248,7 @@ end
 
 ##### CONVERSION #####
 
-# Allows an interval to be converted to a scalar when the set contained by the interval only
-# contains a single element.
-function Base.convert(::Type{T}, interval::Interval{T}) where T
+function Base.only(interval::Interval)
```
omus, line 251: Also needs a proper deprecation

omus (commented):
With a few changes and a rebase to get the CI working I think we can get this merged

pabloferz:
The failing tests are unrelated, but I' not sure about the docs failure.

src/interval.jl
```diff
@@ -233,20 +233,32 @@ function Base.maximum(interval::AbstractInterval{T,L,Open}; increment=nothing) w
...
+##### CONVERSION #####
 
+function Base.convert(::Type{T}, interval::Interval{T}) where {T}
+    depwarn(
+        "`convert(::Type{T}, interval::Interval{T})` is deprecated, " *
+        "use `only(interval)` instead.",
+        :convert,
+    )
+    return only(interval)
+end
```
omus, line 259: Move this to `src/deprecated.jl`. You may be able to define this as:
```julia
@deprecated (convert(::Type{T}, interval::Interval{T}) where T) only(interval)
```
Also, take note there are some deprecations near the top that need to be updated to use `only`

omus (commented):
Only some minor things left here

src/interval.jl
pabloferz: I don't think this will work, if we do, it creates a `Intervals.convert` that overshadows `Base.convert`, and we cannot use `Base.convert` within `@deprecated`.

src/interval.jl
```diff
@@ -233,20 +233,32 @@ function Base.maximum(interval::AbstractInterval{T,L,Open}; increment=nothing) w
...
+"""
+    only(interval::AbstractInterval)
 
-# Allows an interval to be converted to a scalar when the set contained by the interval only
-# contains a single element.
-function Base.convert(::Type{T}, interval::Interval{T}) where T
+Returns the only element of a closed interval with coinciding endpoints (throws a
+DomainError otherwise).
+"""
+function Base.only(interval::Interval)
```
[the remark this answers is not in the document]
pabloferz: I went for the generic definition here.

Issue #146, opened 2020-10-08
Implement rounding for intervals


omus:
We should implement `floor`, `ceil`, and `round` for intervals. Unfortunately there are some open questions around how to support rounding on intervals:

1. Which endpoint should be rounded? Left, right, both?
2. Should the span between the endpoints remain consistent?

Let's look at some examples of possible implementations. Let's start with rounding both endpoints:
```julia
julia> round(Interval(0.6, 1.6))  # Rounds both endpoints up, span remains the same
Interval{Float64,Closed,Closed}(1.0, 2.0)

julia> round(Interval(0.4, 1.4))  # Rounds both endpoints down, span remains the same
Interval{Float64,Closed,Closed}(0.0, 1.0)

julia> round(Interval(0.6, 1.4))  # Rounds left up and right down, span differs
Interval{Float64,Closed,Closed}(1.0, 1.0)

julia> round(Interval(0.5, 1.5))  # Note: Behaviour due to default of `RoundNearest`
Interval{Float64,Closed,Closed}(0.0, 2.0)

julia> round(Interval(1.5, 2.5))
Interval{Float64,Closed,Closed}(2.0, 2.0)
```

From looking at these examples rounding one endpoint up and one down definitely doesn't seem right.
Now let's take a look at some other where we floor both endpoints:

```julia
julia> floor(Interval(0.6, 1.4))  # Span was 0.8, after flooring is 1.0
Interval{Float64,Closed,Closed}(0.0, 1.0)

julia> floor(Interval(0.1, 1.9))  # Span was 1.8, after flooring is 1.0
Interval{Float64,Closed,Closed}(0.0, 1.0)
```
After seeing flooring both endpoints in practice I think that the span should probably remain the same after the operation. 

This would mean that we should probably only perform round, floor, and ceil on a specified endpoint. Effectively, this just shifts the interval and the span always remains the same. Such an implementation also seems to work well for `AnchoredInterval` however we may want to have the option to round based on the anchor (which may be the left or right endpoint).

omus:
I have [previously made an attempt to implement rounding](https://github.com/invenia/Intervals.jl/pull/65) but ran into these problems.

omus:
For unbounded intervals if you attempt to floor/ceil/round the unbounded endpoint I would expect the interval to remain unchanged. Operating on the bounded endpoint would modify the bounded endpoint but leave the unbounded endpoint as unbounded. This would be consistent with how scalar arithmetic works with intervals.

PR #147, opened 2020-10-09
Implement interval rounding

nicoleepp:
Supersedes https://github.com/invenia/Intervals.jl/pull/65

Closes https://github.com/invenia/Intervals.jl/issues/146

~Still needs documentation additions but should be ready for initial review~ 
Now has documentation


test/interval.jl
```diff
@@ -887,4 +887,119 @@
...
+        @test floor(Interval{Unbounded, Closed}(nothing, 1.0), on=RightEndpoint) == Interval{Unbounded, Closed}(nothing, 1.0)
+        @test floor(Interval{Unbounded, Closed}(nothing, 1.5), on=RightEndpoint) == Interval{Unbounded, Closed}(nothing, 1.0)
+        @test floor(Interval{Unbounded, Unbounded}(nothing, nothing), on=RightEndpoint) == Interval{Unbounded, Unbounded}(nothing, nothing)
+    end
+
+    @testset "ceil" begin
+        @test_throws UndefKeywordError ceil(Interval(0.0, 1.0))
+
+        @test ceil(Interval(0.0, 1.0), on=LeftEndpoint) == Interval(0.0, 1.0)
+        @test ceil(Interval(0.5, 1.0), on=LeftEndpoint) == Interval(1.0, 1.5)
```
omus, line 933: I'm not sure `on` is the right keyword as it makes me think only the left would be changed. I'll think about a better name

test/interval.jl
```diff
@@ -887,4 +887,119 @@
...
+
+        @test floor(Interval(0.0, 1.0), on=LeftEndpoint) == Interval(0.0, 1.0)
+        @test floor(Interval(0.5, 1.0), on=LeftEndpoint) == Interval(0.0, 0.5)
+        @test floor(Interval(0.0, 1.5), on=LeftEndpoint) == Interval(0.0, 1.5)
+        @test floor(Interval(0.5, 1.5), on=LeftEndpoint) == Interval(0.0, 1.0)
+
+        @test floor(Interval(0.0, 1.0), on=RightEndpoint) == Interval(0.0, 1.0)
+        @test floor(Interval(0.5, 1.0), on=RightEndpoint) == Interval(0.5, 1.0)
+        @test floor(Interval(0.0, 1.5), on=RightEndpoint) == Interval(-0.5, 1.0)
+        @test floor(Interval(0.5, 1.5), on=RightEndpoint) == Interval(0.0, 1.0)
```
omus, line 902: We should have tests for attempting to use `AnchoredEndpoint` with an `Interval`

src/interval.jl
```diff
@@ -478,6 +478,42 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
...
+                    left = RightEndpoint{T, L}(right.endpoint - span(interval))
+                end
+            end
+        else
+            throw(ArgumentError("Unhandled `on` type: $on"))
+        end
+
+        return Interval(left, right)
+    end
+end
```
omus, line 515: This implementation pre-dates unbounded intervals and I think we can probably write this a little cleaner

omus: Possibly you could do the real work in and internal function so `on` can take place in dispatch:
```julia
const RoundingFunctionTypes = Union{typeof(floor), typeof(ceil), typeof(round)}

function Base._round(f::RoundingFunctionTypes, interval::Interval{T,L,R}, on::LeftEndpoint, args...) where {T, L <: Bounded, R <: Bounded}
    left_val = f(first(interval), args...)
    return Interval{T,L,R}(left_val, left_val + span(interval))
end
```

src/Intervals.jl
```diff
@@ -42,6 +42,9 @@ export Bound,
        AbstractInterval,
        Interval,
        AnchoredInterval,
+       LeftEndpoint,
+       RightEndpoint,
+       AnchorEndpoint,
```
omus, line 47: I'm less certain about exporting these. Maybe we should just use symbols `:left`, `:right`, and `:anchor` for the initial version of rounding

src/interval.jl
omus: If you switch to using `Symbol` for the `on` keyword you can use `Direction{:left}` or `Val{:left}` to still support dispatch

test/interval.jl
omus: Thinking about it more I think this is just my personal bias and `on` is probably correct

src/Intervals.jl
nicoleepp: The deprecations mean they were being exported anyway (except for AnchorEndpoint)

fchorney (approved):
This looks good to me.

src/endpoint.jl
```diff
@@ -24,10 +24,54 @@ struct Endpoint{T, D, B <: Bound}
...
 
+
 Endpoint{T,D,B}(ep) where {T, D, B <: Bounded} = Endpoint{T,D,B}(convert(T, ep))
 
-const LeftEndpoint{T,B} = Endpoint{T, Left, B} where {T,B}
-const RightEndpoint{T,B} = Endpoint{T, Right, B} where {T,B}
+"""
+    LeftEndpoint
+
+`LeftEndpoint` represents the lesser endpoint of an `AbstractInterval`. Useful for comparing
```
morris25, line 33: I don't think we need to restate the name here.

morris25 (approved):
LGTM

test/interval.jl
```diff
@@ -887,4 +887,140 @@
...
+        @test floor(Interval(0.0, 1.5); on=:left) == Interval(0.0, 1.5)
+        @test floor(Interval(0.5, 1.5); on=:left) == Interval(0.0, 1.0)
+
+        @test floor(Interval(0.0, 1.0); on=:right) == Interval(0.0, 1.0)
+        @test floor(Interval(0.5, 1.0); on=:right) == Interval(0.5, 1.0)
+        @test floor(Interval(0.0, 1.5); on=:right) == Interval(-0.5, 1.0)
+        @test floor(Interval(0.5, 1.5); on=:right) == Interval(0.0, 1.0)
+
+        # :anchor is only usable with AnchoredIntervals
+        @test_throws MethodError floor(Interval(0.0, 1.0); on=:anchor)
```
mattBrzezinski, line 909: Is it worth while to implement this method which then throws an explicit error saying you cannot anchor on a non-AnchoredInterval?

I ask cause it may seem at first glance like this is missing an implementation if you attempt it.

mattBrzezinski (approved):
LGTM, just had one question which might lead to a change.

src/endpoint.jl
```diff
@@ -24,10 +24,54 @@ struct Endpoint{T, D, B <: Bound}
...
+julia> RightEndpoint{Closed}(1.0)
+Intervals.Endpoint{Float64,Intervals.Direction{:Right}(),Closed}(1.0)
+
+julia> RightEndpoint{Integer, Closed}(1.0)
+Intervals.Endpoint{Integer,Intervals.Direction{:Right}(),Closed}(1)
+```
+
+See also: [`LeftEndpoint`](@ref)
+"""
+const RightEndpoint{T,B} = Endpoint{T, Right, B} where {T,B <: Bound}
```
omus, line 74: I wouldn't mind the docstrings being in a set PR

omus (commented):
Looks good. I think there could be some further iteration on the implementation but I'm good to proceed

PR #151, opened 2020-11-27
Use GitHub Actions

fchorney:
Ok GitHub Actions:

CI.yml: 
 - Runs on a cron, 2 AM UTC every day
 - Runs on push for master branch and tags
 - Runs on pull requests

 - Runs Julia 1 (x64) on Linux, MacOS, and Windows
 - Runs 1.0.5 (Linux x64)
 - Runs 1.3 (Linux x64)
 - Runs 1 (Linux x86)

 - Will notify Slack on failure (one slack message for all failures, uses workflow conclusion)

JuliaNighly.yml:
 - Runs on a cron, 2 AM UTC every day

 - Runs Julia Nightly (Linux, x64)

 - Will not notify Slack 

General updates to ComparHelper.yml and TagBot.yml using recommended settings.


nicoleepp:
> There is also the issue that you can't "allow failure" yet, so we will be spammed with nightly failures for the time being.

Do we want to run nightly on master branches if that is the case? Maybe we could make it a manual trigger? Or is there a plan to allow_failures soon?

.github/workflows/CIMaster.yml
```diff
@@ -0,0 +1,92 @@
...
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
+        arch:
+          - x64
+        include:
+          # Add a 1.0 job just to make sure we still support it
+          - os: ubuntu-latest
+            version: 1.0.5
```
nicoleepp, line 27: Would 1.0 here just grab the latest 1.0.x version?

fchorney: I think I read somewhere that if you do `1.0` it strips out the `.0` for some reason and grabs the latest `1.x` package. Found the link https://github.com/invenia/JLSO.jl/pull/81#discussion_r516857524

fchorney:
> > There is also the issue that you can't "allow failure" yet, so we will be spammed with nightly failures for the time being.
> 
> Do we want to run nightly on master branches if that is the case? Maybe we could make it a manual trigger? Or is there a plan to allow_failures soon?

I think they are going to support it at some point, but there isn't necessarily a guaranteed time when it may come maybe. I don't think there are manual triggers yet either. https://github.com/actions/toolkit/issues/399

.github/workflows/CI.yml
```diff
@@ -0,0 +1,82 @@
+name: CI
+# Run on tags or any pull request
+on:
+  push:
```
rofinn, line 4: Needs a schedule

.github/workflows/CI.yml
```diff
@@ -0,0 +1,82 @@
...
+    tags: ["*"]
+  pull_request:
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.version == 'nightly' }}
```
rofinn, line 11: This doesn't currently work the way we want so maybe we should just stop testing on julia nightly unless we're okay with spam.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,82 @@
...
+      fail-fast: false
+      matrix:
+        version:
+          - 1
+        os:
+          - ubuntu-latest
+          - macOS-latest
+          - windows-latest
+        arch:
+          - x64
```
rofinn, line 22: Intervals should maybe work on `x86`?

fchorney: The doctests fail.  Tried it

.github/workflows/CI.yml
fchorney: yeah continue on error is only used when fail fast is set to true. Potentially we just don't test master on nightly at all until the allow failure is fixed

.github/workflows/CIMaster.yml
```diff
@@ -0,0 +1,92 @@
+name: CI
```
rofinn, line 1: So we basically need to have separate CI files just for the schedule and slack notifications? :(

.github/workflows/CI.yml
rofinn: Maybe make an issue? We can probably just add a condition for only running the doctests on 64-bit.

fchorney:
I'm not sure why the CI isn't currently running. It's not telling me that the CI script is broken or anything. hmm.

.github/workflows/CI.yml
fchorney: hmm so the doctests run with the regular julia tests. Since those would be loaded up on the x86 machine I don't think this is possible? https://github.com/invenia/Intervals.jl/pull/151/checks?check_run_id=1465597418 

Currently it looks like the issues are with doctests saying something will be `Int64` but they are `Int32` on the x86 build which makes sense. Would a fix be to just change those to `Int`?

.github/workflows/TagBot.yml
```diff
@@ -9,3 +9,4 @@ jobs:
       - uses: JuliaRegistries/TagBot@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
+          ssh: ${{ secrets.DOCUMENTER_KEY }}
```
nicoleepp, line 12: Why is this needed?

mattBrzezinski: Not familiar w/ Documenter but is it not needed for pushing changes?

> https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#travis-ssh

rofinn: It's needed if you're also using a documenter ssh key for another action 
https://github.com/JuliaRegistries/TagBot#setup

.github/workflows/CI.yml
rofinn: I think the repl output will still show it as `Int64` though. I'd just condition the `doctest` call on the architecture.

mattBrzezinski:
GH was being weird, so my comments might be out of date.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,82 @@
+name: CI
+# Run on tags or any pull request
+on:
```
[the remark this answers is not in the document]
rofinn: Isn't that what the Nightly.yml file is for?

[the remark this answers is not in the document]
mattBrzezinski: Idk what happened to these comments, I wrote them before Fernando made changes and thought I submitted them. But then they couldn't be deleted?

Ignore this.

.github/workflows/CIMaster.yml
mattBrzezinski: https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851

Looks like this has been a thing for a while... annoying.

.github/workflows/CI.yml
fchorney: oh like from within the julia `runtests.jl` file?

fchorney:
At this point I think we just have to deal with failing Julia-Nightly tests on branches/tags/master/etc, and just let them fail, unless we want to disable Julia-Nightly tests all together for now.

.github/workflows/CompatHelper.yml
```diff
@@ -0,0 +1,16 @@
+name: CompatHelper
+on:
+  schedule:
+    - cron: '00 00 * * *'  # Everyday at midnight
```
mattBrzezinski, line 4: I don't think you need `00` here?

```suggestion
    - cron: '0 0 * * *'  # Everyday at midnight
```

mattBrzezinski (approved):

.github/workflows/CompatHelper.yml
fchorney: yeah I think you're right. I just copied this file from JLSO I think

PR #152, opened 2020-12-01
Define ≈ operator for Intervals

### LINKED ISSUE #150: Define ≈ operator for Intervals
```
julia> i1 = Interval{Float64, Open, Closed}(1, 2)
Interval{Float64,Open,Closed}(1.0, 2.0)

julia> i2 = Interval{Float64, Open, Closed}(1, 2 + 1e-10)
Interval{Float64,Open,Closed}(1.0, 2.00001)

julia> i1 ≈ i2
ERROR: MethodError: no method matching isapprox(::Interval{Float64,Open,Closed}, ::Interval{Float64,Open,Closed})
```

How would this behaviour be defined?

- The lower and upper bounds would have to be ≈ equal separately.

- The closed/open distinction would be ignored.

- If the underlying types don't have ≈ defined an error is thrown?

gxyd:
Fixes #150 

Still need to think about tests for `Unbound` intervals.

### DIFF
--- a/src/interval.jl
+++ b/src/interval.jl
@@ -310,6 +310,24 @@ function Base.isequal(a::AbstractInterval, b::AbstractInterval)
     return le && re
 end
 
+function Base.:(≈)(a::AbstractInterval, b::AbstractInterval)
+    first_a, last_a = first(a), last(a)
+    first_b, last_b = first(b), last(b)
+    # deal with unbounded intervals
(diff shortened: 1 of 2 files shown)


gxyd:
I didn't expect the below behavior in Julia:
```jl
julia> (1 + 1e-10) ≈ 1
true
julia> 1e-10 ≈ 0
false
```

Is the second answer expected to be `false`?

PR #154, opened 2021-01-13
Clean up CI and fix Slack Notification

fchorney:


.github/workflows/CI.yml
```diff
@@ -65,7 +62,7 @@ jobs:
     name: Notify Slack Failure
     needs: test
     runs-on: ubuntu-latest
-    if: github.event == 'schedule'
+    if: always() && github.event_name == 'schedule'
```
omus, line 65: `always()` seems strange to me

omus (approved):

.github/workflows/CI.yml
omus: I see: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-not-requiring-dependent-jobs-to-be-successful

fchorney: so that's needed because of the tests fail then this slack job doesn't run. This job has `needs: test` which github actions seems to think means `only run if the dependencies pass`. Thus `always()` says `run after "test" even if "test" fails", and then we further limit it to only running if we're on the scheduled run.

PR #155, opened 2021-01-18
Fix CI Tests and Artifact Cache Issues

fchorney:
Use CLI script to automatically replace the 1.3 test with a 1.5 test (if necessary), as well as use new cache steps to cache the julia build artifacts

### DIFF
--- a/.github/workflows/JuliaNightly.yml
+++ b/.github/workflows/JuliaNightly.yml
@@ -13,16 +13,14 @@ jobs:
         with:
           version: nightly
           arch: x64
-      - uses: actions/cache@v1
+      - uses: actions/cache@v2
         env:
-          cache-name: cache-artifacts
+          cache-name: julia-nightly-cache-artifacts
         with:
           path: ~/.julia/artifacts
-          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          key: ${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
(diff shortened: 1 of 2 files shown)


nicoleepp (approved):

.github/workflows/JuliaNightly.yml
```diff
@@ -13,16 +13,14 @@ jobs:
...
           arch: x64
-      - uses: actions/cache@v1
+      - uses: actions/cache@v2
         env:
-          cache-name: cache-artifacts
+          cache-name: julia-nightly-cache-artifacts
         with:
           path: ~/.julia/artifacts
-          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
+          key: ${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
```
omus, line 21: Just curious why we don't keep these the same between CI.yml and here?

omus (approved):

.github/workflows/JuliaNightly.yml
fchorney: Hmm I might have incorrectly assumed since they were in completely separate files/pipelines that the caches wouldn't overlap, this I chose to minimize the key on this one since the `runner.os` is always the same for this test.

fchorney: I don't know that this confirms if caches work across all ci configs in a project specifically https://github.com/actions/cache but thinking about it more, it might just be project wide

PR #156, opened 2021-02-21
Register the Interval structs with Arrow.jl via Requires.jl

rofinn:


rofinn:
I guess this would require us to drop testing on Julia 1.0 since Arrow requires 1.3.

.github/workflows/CI.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - "1.0"  # LTS
+          - "1.3"  # Oldest release supported by Arrow.jl
```
omus, line 18: Need to either set the package Julia company to 1.3 or we should add `VERSION` checks and continue to support Julia 1.0 but disable Arrow integration

test/arrow.jl
```diff
@@ -0,0 +1,15 @@
...
+    zdt_start = ZonedDateTime(2016, 8, 11, 1, tz"America/Winnipeg")
+    zdt_end = ZonedDateTime(2016, 8, 12, 0, tz"America/Winnipeg")
+    table = (time = HE.(zdt_start:Hour(1):zdt_end), val = rand(24))
+
+    # Just test that we can save and load our table type without the time column being
+    # converted to a NamedTuple
+    mktempdir() do d
+        file = joinpath(d, "data.arrow")
+        Arrow.write(file, table)
+        loaded = Arrow.Table(file)
```
omus, line 11: Adding an `isa` test would be nice

test/arrow.jl
```diff
@@ -0,0 +1,15 @@
...
+        @test loaded.time == table.time
+        @test loaded.val ≈ table.val
```
omus, line 13: Why is this approximate?

rofinn: Cause the values are floats. Could have opted for ints I guess?

.github/workflows/CI.yml
rofinn: The thing is that this package will continue working on 1.0 and we can't use version checks because the test dependencies will just fail to install before we even try to load our code.

omus: Ah, I see. I dislike not being able to test on the older version supported by the package but I understand the change now.

test/arrow.jl
omus: Makes sense. If you're making any other changes I'd switch this to ints.

test/runtests.jl
```diff
@@ -1,3 +1,4 @@
+using Arrow
```
omus, line 1:
```suggestion
VERSION >= v"1.3" && using Arrow
```

test/runtests.jl
```diff
@@ -19,6 +20,7 @@ include("test_utils.jl")
     include("anchoredinterval.jl")
     include("comparisons.jl")
     include("plotting.jl")
+    include("arrow.jl")
```
omus, line 23:
```suggestion
    VERSION >= v"1.3" && include("arrow.jl")
```

test/runtests.jl
omus: Would allow the tests to work if you work on Julia 1.0 if you work around Arrow.jl requiring 1.3

rofinn: It still wouldn't resolve to hit this anyways cause `Pkg.test` would error trying to resolve the extra dependency. If `Pkg` allowed you to conditionally install packages then this would work.

test/runtests.jl
rofinn: Again, this isn't really helpful because we still need to declare Arrow.jl in our Project.toml.

test/runtests.jl
omus: I know. This just helps for when someone manually removes `Arrow` from the test dependency list and runs the tests.

rofinn: 👌🏻

test/runtests.jl
rofinn: Okay, included a comment at the top of the file to explain why this is here. Since it probably isn't very intuitive why that would be desired.

test/runtests.jl
```diff
@@ -1,3 +1,6 @@
+# NOTE: v1.3 checks are for if you want to test on pre-1.3 w/o the Arrow.jl dependency
+VERSION > v"1.3" && using Arrow
```
omus, line 2:
```suggestion
VERSION >= v"1.3" && using Arrow
```

omus (approved):
Should be good with the last correction

omus:
> One thing to consider with this addition is that Arrow.jl doesn't handle type changes very well.

I'm going to run a couple experiments with this today and either merge as-is or make some changes.

omus:
Looks like the parametric information isn't being stored which is problematic:
```julia
julia> using Arrow

julia> using Dates, Intervals, TimeZones

julia> table = (interval = [
           Interval{Closed,Closed}(1,2),
           Interval{Open,Closed}(1,2),
           Interval{Closed,Unbounded}(1,nothing),
       ],)
(interval = Interval{Int64,L,R} where R<:Bound where L<:Bound[Interval{Int64,Closed,Closed}(1, 2), Interval{Int64,Open,Closed}(1, 2), Interval{Int64,Closed,Unbounded}(1, nothing)],)

julia> Arrow.write("data.arrow", table)

julia> loaded = Arrow.Table("data.arrow");

julia> loaded.interval
3-element Arrow.Struct{Interval{Int64,L,R} where R<:Bound where L<:Bound,Tuple{Arrow.Primitive{Int64,Array{Int64,1}},Arrow.Primitive{Int64,Array{Int64,1}}}}:
 Interval{Int64,Closed,Closed}(1, 2)
 Interval{Int64,Closed,Closed}(1, 2)
 Interval{Int64,Closed,Closed}(1, 1)
```

omus:
I've made issues with Arrow for the two problems I've discovered:
- https://github.com/JuliaData/Arrow.jl/issues/134
- https://github.com/JuliaData/Arrow.jl/issues/135

omus (changes requested):
Need to address Arrow issues

Issue #163, opened 2021-04-09
MethodError trying to `print` a `AnchoredInterval{Week(-1)}`


nickrobinson251:
The behaviour for printing a `AnchoredInterval{Day(-1)}` is:
```julia
julia> dt = DateTime(2016, 8, 11, 2)
2016-08-11T02:00:00

julia> AnchoredInterval{Day(-1)}(DateTime(dt))
AnchoredInterval{Day(-1), DateTime, Open, Closed}(DateTime("2016-08-11T02:00:00"))

julia> print(ans)
(DE 2016-08-11 02:00:00]
```

And this style of `print` is available for `AnchoredInterval{T(-1)}` if `T` is `Day`, `Month`, Year` or  `Millisecond`, `Second`, `Minute`, `Hour`

 But _errors_ for `Week`, `Quarter` (new in Julia v1.6), or `Nanosecond`, `Microsecond`
    
E.g. for `AnchoredInterval{Week(-1)}`
```julia
julia> dt = DateTime(2016, 8, 11, 2)
2016-08-11T02:00:00

julia> AnchoredInterval{Week(-1)}(DateTime(dt))
AnchoredInterval{Week(-1), DateTime, Open, Closed}(DateTime("2016-08-11T02:00:00"))

julia> print(ans)
ERROR: MethodError: no method matching prefix(::Type{Week})
Closest candidates are:
  prefix(::Period) at /Users/nick/.julia/packages/Intervals/ua9cq/src/description.jl:58
  prefix(::Type{Year}) at /Users/nick/.julia/packages/Intervals/ua9cq/src/description.jl:62
  prefix(::Type{Month}) at /Users/nick/.julia/packages/Intervals/ua9cq/src/description.jl:63
  ...
Stacktrace:
 [1] prefix(p::Week)
   @ Intervals ~/.julia/packages/Intervals/ua9cq/src/description.jl:59
 [2] description(dt::DateTime, p::Week, suffix::String)
   @ Intervals ~/.julia/packages/Intervals/ua9cq/src/description.jl:41
 [3] description(interval::AnchoredInterval{Week(-1), DateTime, Open, Closed}, s::String)
   @ Intervals ~/.julia/packages/Intervals/ua9cq/src/description.jl:4
 [4] description(i::AnchoredInterval{Week(-1), DateTime, Open, Closed})
   @ Intervals ~/.julia/packages/Intervals/ua9cq/src/description.jl:1
[...]

PR #164, opened 2021-04-20
Add `prefix` methods for `Week` and `Quarter` Types.  

Ianlmgoddard:
partially tackles #163 

I haven't included prefixes for `Microsecond` or `Nanosecond` because as far as I can tell, `DateTime`s are only defined up to millisecond precision (see https://github.com/JuliaLang/julia/issues/31525). I wasn't sure if we wanted to display `DateTime` Intervals with greater precision than what is available for `DateTime`s.

I also haven't added tests for `Quarter` because it's not clear to me how to get around the fact that this type was only introduced in Julia v1.6. Any advice there would be appreciated.

### DIFF
--- a/src/description.jl
+++ b/src/description.jl
@@ -60,7 +60,6 @@ function prefix(p::Period)
 end
 
 prefix(::Type{Year}) = "Y"
-prefix(::Type{Quarter}) = "Q"
 prefix(::Type{Month}) = "Mo"
 prefix(::Type{Week}) = "W"
 prefix(::Type{Day}) = "D"
@@ -69,5 +68,10 @@ prefix(::Type{Minute}) = "M"
 prefix(::Type{Second}) = "S"
 prefix(::Type{Millisecond}) = "ms"
 
-# informative error if no prefix is defined for the givne type.
+# `Quarter` defined after Julia1.6.
+if VERSION >= v"1.6"
+    prefix(::Type{Quarter}) = "Q"
+end
+
+# informative error if no prefix is defined for the given type.
 prefix(T::Type{<:Period}) = error("A prefix for period $T has not yet been defined")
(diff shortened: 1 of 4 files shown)


omus:
> I also haven't added tests for `Quarter` because it's not clear to me how to get around the fact that this type was only introduced in Julia v1.6. Any advice there would be appreciated.

Use conditional blocks like:
```julia
if VERSION >= v"1.6"
    ...
end
```
Which allows you to only run code if the Julia version is above or equal to 1.6.0. 

It shouldn't matter here but if you use `@static` before the `if` Julia can do the check at compile time allowing you to write unsupported syntax.

src/description.jl
```diff
@@ -61,8 +61,17 @@ end
...
 prefix(::Type{Year}) = "Y"
 prefix(::Type{Month}) = "Mo"
+prefix(::Type{Week}) = "W"
 prefix(::Type{Day}) = "D"
 prefix(::Type{Hour}) = "H"
 prefix(::Type{Minute}) = "M"
 prefix(::Type{Second}) = "S"
 prefix(::Type{Millisecond}) = "ms"
+
+# `Quarter` defined after Julia1.6.
```
LilithHafner, line 71:
```suggestion
# `Quarter` defined in Julia v1.6.
```

src/description.jl
```diff
@@ -61,8 +61,17 @@ end
...
+if VERSION >= v"1.6"
+    prefix(::Type{Quarter}) = "Q"
+end
+
+# informative error if no prefix is defined for the given type.
+prefix(T::Type{<:Period}) = error("A prefix for period $T has not yet been defined")
```
LilithHafner, line 77: If we'd like it to "just work" instead of throwing,
```suggestion

# Full name if no prefix is defined.
prefix(T::Type{<:Period}) = string(T)
```

LilithHafner (commented):
Thanks for fixing these missing prefixes!

Issue #159, opened 2021-03-25
Plotting CIs throws several warnings 


AlexRobson:
In the present version of plots, I am hitting [this warning ](https://github.com/JuliaPlots/Plots.jl/blob/master/src/utils.jl#L104)which is causing a lot of spam around plots. 

At some point, it looks like Plots changed from using `NaN` segments to using a vector of coordinates. This breaks the assumption [here](https://github.com/invenia/Intervals.jl/blob/master/src/plotting.jl#L16). Under the hood, I believe this was converting the `y` series into something like `[1.0, 1.0, NaN, 1.0, 1.0, NaN, 1.0, 1.0, NaN]`...etc. Now however, we want to do `[[1.0, 1.0], [1.0, 1.0], [1.0, 1.0]]`

This looks like it will fix it:

```
@recipe function f(xs::AbstractVector{<:AbstractInterval{T,L,R}}, ys) where {T, L <: Bounded, R <: Bounded}
    new_xs = Vector{T}[]
    new_ys = Vector{Any}[]
    markers = Symbol[]
    for (x, y) in zip(xs, ys)
        # To cause line to not be connected, need to divide individually separated segments into separate vectors
        append!(new_xs, [[first(x), last(x)]])
        append!(new_ys, [[y, y]])
    end
    append!(markers, interval_markers(first(xs)))
    
    # Work around GR bug that shows :none as a marker
    # TODO: remove once https://github.com/jheinen/GR.jl/issues/295  is fixed
    markeralpha := [x == :none ? 0 : 1 for x in markers]

    markershape := markers
    seriestype  := :path  # always a path, even in a scatter plot

    new_xs, new_ys
    
end
```

Can make this into an MR, but this would break older versions of Plots. Can keep both. Looks like behaviour changes slightly too, as now it iterates through the colour map (each line is a different colour). 

Related: https://github.com/invenia/Intervals.jl/issues/134

laborg:
I've came across this issue too, and wonder why a vector of intervals is plotted as a single path. Currently the plotting of Intervals is also broken for vectors of Intervals where the bounds are mixed (e.g. [Interval{Open,Closed}(1,2),Interval{Closed,Closed}(2,3)] is excluded from the signature. For this to work the where clause needs to be moved.

PR #165, opened 2021-04-20
Fix plotting in Plots > v.1.11

AlexRobson:
Fixes #https://github.com/invenia/Intervals.jl/issues/159

Plots changes the way it handles line segments. Instead of `NaN` separated, it uses a vector of points (see issue). 

Should we keep support for Plots < v1.11

### DIFF
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -5,24 +5,25 @@ interval_marker(x::Type{Open}) = :none
 
 # TODO: Add support for plotting unbounded intervals
 @recipe function f(xs::AbstractVector{<:AbstractInterval{T,L,R}}, ys) where {T, L <: Bounded, R <: Bounded}
-    new_xs = T[]
-    new_ys = []
+    new_xs = Vector{T}[]
+    new_ys = Vector{Any}[]
     markers = Symbol[]
     for (x, y) in zip(xs, ys)
-        # To cause line to not be connected, need to add a breaker point with
-        # NaN in one of the coordinates. We put that in `new_ys` as that is probably a
-        # float. where-as new_xs is potentially a DateTime etc that would not accept NaN
-        append!(new_xs, [first(x), last(x), last(x)])
-        append!(new_ys, [y, y, NaN])
-        append!(markers, interval_markers(x))
-        push!(markers, :none)
+        # To cause line to not be connected, need to divide individually separated segments into separate vectors
+        # This is as of v1.11 in Plots.
+        # See https://github.com/JuliaPlots/Plots.jl/blob/master/src/utils.jl#L104
+        append!(new_xs, [[first(x), last(x)]])
+        append!(new_ys, [[y, y]])
     end
-
+    append!(markers, interval_markers(first(xs)))
+    
     # Work around GR bug that shows :none as a marker
     # TODO: remove once https://github.com/jheinen/GR.jl/issues/295  is fixed
     markeralpha := [x == :none ? 0 : 1 for x in markers]
 
     markershape := markers
     seriestype  := :path  # always a path, even in a scatter plot
+
     new_xs, new_ys
+    
 end
--- a/src/plotting.jl
+++ b/src/plotting.jl
@@ -23,7 +23,5 @@ interval_marker(x::Type{Open}) = :none
 
     markershape := markers
     seriestype  := :path  # always a path, even in a scatter plot
-
     new_xs, new_ys
-    
 end


LilithHafner:
> Should we keep support for Plots < v1.11

On April 20, when @AlexRobson created this PR I would have said yes. 

Now I'd say no. Plots v1.11 has been out since March 2021 and plots is now on v1.29.1. I think it is okay to drop support for 15 month old Plots. If a project has Plots pinned to before 1.11, then they will end up having Intervals pinned to 1.7.1 (or whenever this merges) which seems perfectly acceptable to me.

oxinabox (commented):
Lets not worry about supporting Plots pre-1.11
we can't directly restrict that.
we can restrict to RecipeBase 1.0, or even 1.2, which is the minimum for Plots v1.24
We should drop support for RecipesBase 0.8 and 0.9

so lets make the changes to the Project.toml, including bumping the version number
and release this yes?

I assume this is manually tested, and i think automatically testing this more than we currently are isn't work the effort.
Though perhaps the current automtic tests needs updating?
I am not sure what is up with them since they expired

AlexRobson:
I'm pikcing this back up. Just working through it now. I should have an update shortly.

AlexRobson:
Looking at this again, I recall the reason I hit a block on it, and why it shouldn't go in as-is. There are two reasons for this, one major and one minor. The major reason is that this seems to break with DateTime Intervals. As-is, this will throw an error in one of the tests that include datetime intervals: 

```
ERROR: Cannot convert DateTime to series data for plotting
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:33
  [2] _prepare_series_data(x::DateTime)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/OXGmH/src/series.jl:8
  [3] _series_data_vector(x::DateTime, plotattributes::Dict{Symbol, Any})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/OXGmH/src/series.jl:35
  [4] (::RecipesPipeline.var"#27#30"{Dict{Symbol, Any}})(vi::DateTime)
    @ RecipesPipeline ./none:0
  [5] iterate
    @ ./generator.jl:47 [inlined]
  [6] _series_data_vector(v::Vector{DateTime}, plotattributes::Dict{Symbol, Any})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/OXGmH/src/series.jl:50
  [7] #27
    @ ./none:0 [inlined]
  [8] iterate(::Base.Generator{Vector{Vector{DateTime}}, RecipesPipeline.var"#27#30"{Dict{Symbol, Any}}})
    @ Base ./generator.jl:47
  [9] _series_data_vector(v::Vector{Vector{DateTime}}, plotattributes::Dict{Symbol, Any})
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/OXGmH/src/series.jl:50
 [10] macro expansion
    @ ~/.julia/packages/RecipesPipeline/OXGmH/src/series.jl:134 [inlined]
 [11] apply_recipe(plotattributes::AbstractDict{Symbol, Any}, #unused#::Type{RecipesPipeline.SliceIt}, x::Any, y::Any, z::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesBase/qpxEX/src/RecipesBase.jl:289
 [12] _process_userrecipes!(plt::Any, plotattributes::Any, args::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/OXGmH/src/user_recipe.jl:36
 [13] recipe_pipeline!(plt::Any, plotattributes::Any, args::Any)
    @ RecipesPipeline ~/.julia/packages/RecipesPipeline/OXGmH/src/RecipesPipeline.jl:70
 [14] _plot!(plt::Plots.Plot, plotattributes::Any, args::Any)
    @ Plots ~/.julia/packages/Plots/lW9ll/src/plot.jl:209
 [15] plot(::Any, ::Vararg{Any, N} where N; kw::Any)
    @ Plots ~/.julia/packages/Plots/lW9ll/src/plot.jl:91
 [16] plot(::Any, ::Any)
    @ Plots ~/.julia/packages/Plots/lW9ll/src/plot.jl:85
 [17] top-level scope
    @ REPL[41]:1
```

However, I suspect that this may need to be handled in a different package, I'm playing around with the other packages in the stacktrace to track it down (e.g. RecipesPipelines). 

The minor reason is the change in behaviour means that the intervals all get marked as a different series. For non-datetime intervals: 

```
intervals = [Interval{Closed, Closed}(float(x), float(x + 0.5)) for x in 1:11]
plot(intervals, 1:11)

![image](https://user-images.githubusercontent.com/4730712/186109929-164e57c0-edba-455b-a9e5-3ffb6f247ba1.png)
```

oxinabox:
> The minor reason is the change in behaviour means that the intervals all get marked as a different series. For non-datetime intervals:

This is also signficiant since if we had multiple series (like we often do) we won't be able to tell them apparent and the legend will be useless

BioTurboNick:
This may be helpful: if you set `primary = false`, a new series will take the color of the previous series. And if `label = ""`, it's excluded from the legend.

AlexRobson:
> This may be helpful: if you set `primary = false`, a new series will take the color of the previous series. And if `label = ""`, it's excluded from the legend.

Thanks - this fixed that one :) - the last thing for me in this MR then is tracking down why DateTime isn't working.

BioTurboNick:
One thing that may be worth keeping in mind is that until more recent versions of Plots (improved in 1.23.x, fully fixed by 1.30.2), adding many series could get extremely slow because there were several O(n^2) operations over series. I discovered and fixed them, but depending on how many intervals you expect people to be plotting, and if people are on an older version, they could have a negative experience there. >100 or >1000 is where it might get really annoying.

PR #167, opened 2021-07-08
Arrow support

omus:
Replaces https://github.com/invenia/Intervals.jl/pull/156. Changes here include the commits from that PR and uses the updated serialization framework introduces in https://github.com/JuliaData/Arrow.jl/pull/156


omus:
I'm expecting Julia 1.0 to fail as ArrowTypes doesn't support that version of Julia currently

test/arrow.jl
```diff
@@ -0,0 +1,38 @@
...
+    @testset "AnchoredInterval" begin
+        zdt_start = ZonedDateTime(2016, 8, 11, 1, tz"America/Winnipeg")
+        zdt_end = ZonedDateTime(2016, 8, 12, 0, tz"America/Winnipeg")
+        col = HE.(zdt_start:Hour(1):zdt_end)
+
+        table = (; col)
+        t = Arrow.Table(Arrow.tobuffer(table))
+
+        # Arrow.jl converts all Period types into Second
+        @test_broken eltype(t.col) == HourEnding{ZonedDateTime, Open, Closed}
```
omus, line 35: I'll probably end up disabling `AnchoredInterval` support to start with because of this

omus:
> I'm expecting Julia 1.0 to fail as ArrowTypes doesn't support that version of Julia currently

Fixing this in: https://github.com/JuliaData/Arrow.jl/pull/223

src/Intervals.jl
```diff
@@ -1,8 +1,10 @@
 module Intervals
 
+using ArrowTypes: ArrowTypes, JuliaType, arrowname
 using Dates
 using Printf
 using RecipesBase
+using Requires
```
omus, line 7:
```suggestion
```

Project.toml
```diff
@@ -5,20 +5,25 @@ authors = ["Invenia Technical Computing"]
...
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
 Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
 TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 
 [compat]
+ArrowTypes = "1.1"
 Documenter = "0.23, 0.24"
 Infinity = "0.2.3"
 RecipesBase = "0.7, 0.8, 1"
+Requires = "1"
```
omus, line 21:
```suggestion
```

Project.toml
```diff
@@ -5,20 +5,25 @@ authors = ["Invenia Technical Computing"]
 version = "1.5.0"
 
 [deps]
+ArrowTypes = "31f734f8-188a-4ce0-8406-c8a06bd891cd"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
+Requires = "ae029012-a4dd-5104-9daa-d747884805df"
```
omus, line 12:
```suggestion
```

omus (commented):
I can remove depending on Requires.jl

omus:
Should probably wait for https://github.com/JuliaData/Arrow.jl/pull/229 which allows us some additional flexibility with how we encode `Interval` types.

test/arrow.jl
rofinn: We can't explicitly override this behaviour?

omus: I've forgotten a bunch about this PR but it should be possible to encode the period span in the serialized Arrow form. I do remember I was mainly focused on `Interval` support so I think this comment was more about that the implementation currently in place is incomplete.

OTDE (approved):
LGTM!

Project.toml
```diff
@@ -2,16 +2,18 @@ name = "Intervals"
...
 RecipesBase = "3cdcf5f2-1ef4-517c-9805-6587b60abb01"
 Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
 TimeZones = "f269a46b-ccf7-5d73-abea-4c690281aa53"
 
 [compat]
+ArrowTypes = "2.3"
 RecipesBase = "0.7, 0.8, 1"
 TimeZones = "1.7"
-julia = "1.6"
+julia = "1.10"
```
ararslan, line 19: Since you're requiring Julia 1.10, you could have implemented as a package extension instead of a direct dependency.

PR #179, opened 2021-10-21
Multi-interval set operators

haberdashPI:
This defines a set of functions for easily computing intersections, unions etc... of a set of intervals (passed as an array).

cc: @omus, @ericphanson, @kleinschmidt 

This is based off of a closed PR to TimeSpans (https://github.com/beacon-biosignals/TimeSpans.jl/pull/11)


src/interval.jl
```diff
@@ -364,6 +374,257 @@ true
...
+
+# Edge is used to represent the two bounds of an interval seperately. This
+# allows us to sort edges across sets during `mergesets` (see below).
+abstract type AbstractEdge{T}; end
+struct Edge{T} <: AbstractEdge{T}
+    value::T
+    first::Bool
+    index::Int
+    closed::Bool
+end
```
omus, line 387: `Edge` seems to be an `Endpoint` with an `index`. Sorting is also possible via:
```julia
using Intervals: RightEndpoint
sort(intervals, by=RightEndpoint)
```
Let me know if there's more too this than I'm currently seeing

haberdashPI: Oye, I totally missed the `Endpoint` objects 😳

omus: > Oye, I totally missed the `Endpoint` objects 😳

That's why we do code review 😄

omus:
Closing/re-opening to trigger CI

src/endpoint.jl
```diff
@@ -25,6 +26,27 @@ struct Endpoint{T, D, B <: Bound}
...
+
+# when all intervals are half-closed in the same direction (all [a, b) or all
+# (b, a]) set operations over them are closed (e.g. all resulting intervals will
+# be the same type). In these cases we can improve type invariance by using a
+# `HalfOpenEndpoint`, which indicates whether the bounds are left-closed or
+# right-closed, storing the left/right-ness as a flag
+struct DirectionBound{T} end
+const LeftClosed = DirectionBound{:LeftClosed}()
+const RightClosed = DirectionBound{:RightClosed}()
+struct HalfOpenEndpoint{T, B} <: AbstractEndpoint{T}
```
omus, line 40: Intervals are half-open/closed and not endpoints of intervals. This seems to me like maybe you want is a `HalfOpenInterval` type. I'll need to take a closer look into how this is used though

src/interval.jl
```diff
@@ -482,6 +714,103 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
...
+Base.in(x::AbstractInterval, y::AbstractVector{<:AbstractInterval}) = any(yᵢ -> x ∈ yᵢ, y)
+function Base.issetequal(x::AbstractIntervals, y::AbstractIntervals)
+    x, y = mergeclean(x,y)
+    return x == y
+end
+
+asarray(x) = [x]
+asarray(x::AbstractArray) = x
+alength(x) = 1
+alength(x::AbstractArray) = length(x)
```
omus, line 749: I'm guessing you want `length` from an `Interval`? As you're only using this once probably makes more sense to use multiple dispatch on `intersectmap`. Alternatively, defining `length(::AbstractInterval)` would be a better solution but also seems unnecessary

src/interval.jl
```diff
@@ -482,6 +714,103 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
...
+    return isempty(intersect(x, y))
+end
+Base.in(x::AbstractInterval, y::AbstractVector{<:AbstractInterval}) = any(yᵢ -> x ∈ yᵢ, y)
+function Base.issetequal(x::AbstractIntervals, y::AbstractIntervals)
+    x, y = mergeclean(x,y)
+    return x == y
+end
+
+asarray(x) = [x]
+asarray(x::AbstractArray) = x
```
omus, line 747: This function already exists: `vcat` or alternatively `[x;]`

src/interval.jl
```diff
@@ -364,6 +385,206 @@ true
...
+    # objects 
+    # NOTE: since this statement relies only on types, the branches using this
+    # flag should compile away
+    track_endpoint = endpoint_dir_type(x) != endpoint_dir_type(y)
+
+    while !(isempty(x) && isempty(y))
+        # println("-------------------")
+        #=@show=# t = first_is_less(x, y) ? first(x) : first(y)
+        #=@show=# x_isless = first_is_less(x, y)
+        #=@show=# x_equal = first_is_equal(x, y)
```
omus, line 534: Why not just call `iterate` or `Iterators.peel` and then do these checks?

omus (commented):
I didn't quite have time for a full review. I do suspect the half-open concept doesn't need to be a special type and could just be a simple `ishalfopen` function a call. I'll try to dive deep into this tomorrow. I'll mention the tests for this package a pretty extensive so be sure to lean on them

src/interval.jl
haberdashPI: Sometimes one of the iterables is empty; I find this form more readible than introducing the branches to check for `nothing`.

src/endpoint.jl
haberdashPI: The goal is as follows: during the process of "unbunching" intervals into endpoints, and then "bunching" them back into intervals, I want to preserve the fact that the original array has an all left-closed or all right-closed intervals. That's because 
1.) said form leads to type stable arrays  
2.) said form can ignore the checks for edge types in `mergesets` (hence the ).

I guess another approach could be to have a flag and/or special container that denotes the
"all closed in one direction" property, and use the existing Endpoint type in all cases.

haberdashPI:
> I'll try to dive deep into this tomorrow. I'll mention the tests for this package a pretty extensive so be sure to lean on them

I'm still working out a few bits with the larger unit test (outside the tests for sets). Might make sense to wait to review until that happens. I'll mark it as draft again and mark it ready for review when that's resolved.

omus:
> I'm still working out a few bits with the larger unit test (outside the tests for sets). Might make sense to wait to review until that happens. I'll mark it as draft again and mark it ready for review when that's resolved.

Thanks for letting me know. I'll try to keep an eye on this as I think some early course correction could save a bunch of work.

haberdashPI:
Looks like there are a few bits I need to add test coverage for, but it should be in good shape to review.

src/interval.jl
haberdashPI: I guess the other point to make here is that we only want to peel the values when they are "first" (i.e. less than or equal to the value in the other list), so you don't want to peel before that.

haberdashPI:
@omus: Can you help me understand the remaining errors I'm getting in "Julia 1" (I think that is 1.7). It looks like there's something broken about printing that is unrelated to the changes I've made.

In other news, I think this otherwise ready for review.

omus:
> @omus: Can you help me understand the remaining errors I'm getting in "Julia 1" (I think that is 1.7). It looks like there's something broken about printing that is unrelated to the changes I've made.

There was an unrelated CI failure for "Julia 1" so I restarted the CI jobs

> In other news, I think this is otherwise ready for review.

Thanks for letting me know. I don't quite have time to review such a big PR this week but I'll make time next week.

haberdashPI:
> I don't quite have time to review such a big PR this week but I'll make time next week.

Thanks! I understand, it's a bit large. I don't think there's any great urgency on my end; I'm making use of the branch directly where I need it in some analyses

haberdashPI:
Just a small bump here: just wanted to check in about when a review would be possible and/or if I should do something to make reviewing this easier (break it up?).

src/endpoint.jl
omus: The problem with the design as I see it is that endpoints are dealing with the left/right side of the interval without knowing anything about the other endpoint in the interval. I'll be able to speak better to alternatives once I read through the rest of the code.

src/endpoint.jl
```diff
@@ -55,7 +92,16 @@ function Base.hash(x::Endpoint{T,D,B}, h::UInt) where {T,D,B}
...
+function Base.hash(x::HalfOpenEndpoint{T,B}, h::UInt) where {T,B}
+    h = hash(:HalfOpenEndpoint, h)
+    h = hash(B, h)
+    return hash(x.endpoint, h)
+end
+
 Base.broadcastable(e::Endpoint) = Ref(e)
+Base.broadcastable(e::HalfOpenEndpoint) = Ref(e)
+
+offset_value(x::AbstractEndpoint) = isleft(x) ? !isclosed(x) : isclosed(x)
```
omus, line 104: Function name isn't descriptive. At a minimum needs a comment explaining the purpose

src/endpoint.jl
```diff
@@ -76,96 +122,78 @@ Visualizing two contiguous intervals can assist in understanding this logic:
...
-Base.:(==)(a, b::Endpoint) = a == b.endpoint && isclosed(b)
-Base.:(==)(a::Endpoint, b) = b == a
-
-function Base.isless(a, b::LeftEndpoint)
-    return (
-        !isunbounded(b) && (
-            a < b.endpoint ||
-            a == b.endpoint && !isclosed(b)
+Base.:(==)(a::Number, b::Endpoint) = a == b.endpoint && isclosed(b)
+Base.:(==)(a::Endpoint, b::Number) = b == a
```
omus, line 172: Endpoints are not restricted to numeric types. We may need to compare against `Date`s, `DateTime`s, `Char`s etc.

src/interval.jl
```diff
@@ -1,12 +1,13 @@
...
+An `Interval` represents a non-iterable range or span of values (non-interable
+because, unlike a `StepRange`, no step is defined).
 
-An `Interval` can be closed (both `first` and `last` are included in the interval), open
-(neither `first` nor `last` are included), or half-open. This openness is defined by the
-bounds information which is stored as the type parameters `L` and `R`.
+An `Interval` can be closed (both `first` and `last` are included in the
+interval), open (neither `first` nor `last` are included), or half-open. This
+openness is defined by the bounds information which is stored as the type
+parameters `L` and `R`.
```
omus, line 10: This repo follows the Blue style so the max line length for this repo is 92: https://github.com/invenia/BlueStyle#synopsis

src/interval.jl
```diff
@@ -36,6 +37,31 @@ julia> Dates.today() - Dates.Week(1) .. Dates.today()
...
+The `Interval` constructor will compare `first` and `last`; if it finds that
+`first > last`, they will be reversed to ensure that `first < last`. This
+simplifies calls to `in` and `intersect`:
+
+```julia
+julia> i = Interval{Open,Closed}(Date(2016, 8, 11), Date(2013, 2, 13))
+Interval{Date,Closed,Open}(2013-02-13, 2016-08-11)
+```
+
+Note that the bounds are also reversed in this case.
```
omus, line 51: This behaviour is deprecated and in the future will be another way of creating an empty interval

src/interval.jl
```diff
@@ -36,6 +37,31 @@ julia> Dates.today() - Dates.Week(1) .. Dates.today()
...
+
+### Multi-interval set operations
+
+Set operations can also be performed over two pairs interval arrays. These set
+operations take the form of `op(x::Vector{<:Interval}, y::Vector{<:Interval})`
+and intrepret `x` and `y` as representing a set covered by the provided
+intervals. The return value is an array of non-overalping intervals representing
```
omus, line 58:
```suggestion
intervals. The return value is an array of non-overlapping intervals representing
```

src/interval.jl
```diff
@@ -36,6 +37,31 @@ julia> Dates.today() - Dates.Week(1) .. Dates.today()
...
+the result of `op` over these sets. You can also pass a single interval to
+either arugment (e.g. op(x::Interval, y::Vector{<:Interval})).
+
+These mutli-interval set operations currently only support intervals that are
```
omus, line 62:
```suggestion
These multi-interval set operations currently only support intervals that are
```

src/interval.jl
```diff
@@ -351,6 +388,203 @@ true
...
+
+# unbunch: represent one or more intervals as a sequence of endpoints and their indices into the original intervals
+endpoint_type(::Type{<:AbstractInterval{T}}) where T = Endpoint{T}
+endpoint_type(::Type{<:AbstractVector{T}}) where T <: AbstractInterval = endpoint_type(T)
+endpoint_type(x) = endpoint_type(typeof(x))
+
+leftpoint(x::AbstractInterval, ::Type{<:Endpoint}) = LeftEndpoint(x)
+rightpoint(x::AbstractInterval, ::Type{<:Endpoint}) = RightEndpoint(x)
+leftpoint(x::AbstractInterval, ::Type{<:HalfOpenEndpoint}) = LeftHalfOpenEndpoint(x)
+rightpoint(x::AbstractInterval, ::Type{<:HalfOpenEndpoint}) = RightHalfOpenEndpoint(x)
```
omus, line 401: What's the advantage of these over calling the constructors?

src/interval.jl
```diff
@@ -351,6 +388,203 @@ true
 ≫(a, b) = greater_than_disjoint(a, b)
 # ≫̸(a, b) = !≫(a, b)
 
+###### Set-related Helpers #####
+
+# unbunch: represent one or more intervals as a sequence of endpoints and their indices into the original intervals
```
omus, line 393: Comment in the wrong place?

src/interval.jl
```diff
@@ -351,6 +388,203 @@ true
...
+    if enumerate
+        result = mapreduce(((i) -> [(i, leftpoint(intervals[i], T)),
+                                    (i, rightpoint(intervals[i], T))]), vcat, filtered)
+        return sort!(convert(Array{Tuple{Int,T}}, result); by=sortby ∘ last)
+    else
+        result = mapreduce(((i) -> [leftpoint(intervals[i], T), rightpoint(intervals[i], T)]),
+                           vcat, filtered)
+        return sort!(convert(Array{T}, result); by=sortby)
+    end
+end
```
omus, line 424: If the point of a `HalfOpenEndpoint` is to produce a type-stable vector then why not just create two vectors where one contains only `LeftEndpoint`s and the other contains `RightEndpoint`s? This should also eliminate the need for the `enumerate` keyword. For sorting you'll need to use `sortperm` and apply the indices to both vectors.

The only disadvantage I see is that you could only sort on left or right endpoints but not both at the same time. Is that a use case you need for this though?

src/endpoint.jl
omus: > It's not obvious to me, at the moment, how to write that in a type stable way though.

I'd suggest putting code readability over type stability to start with as otherwise you can optimize yourself into a corner

haberdashPI: > The problem with the design as I see it is that endpoints are dealing with the left/right side of the interval without knowing anything about the other endpoint in the interval

My idea was to create a subtype of endpoint that is only used in those cases where you do know something about the other endpoints. I do also feel that there is probably a better design, combing back to this now (after some time away). I'll think it over some more and wait for whatever further comments you have.

src/interval.jl
```diff
@@ -469,6 +714,100 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
...
+    data::E
+end
+offset(x::Endpoint) = EndpointOffset(x)
+function Base.isless(x::EndpointOffset, y::EndpointOffset)
+    if isequal(x.data, y.data)
+        return isless(offset_value(x.data), offset_value(y.data))
+    else
+        return isless(x.data, y.data)
+    end
+end
```
omus, line 760: Code is far more complex then it needs to be. What should be done is this can be turned into a custom `lt` function for `sort` and can be turned into:
```suggestion
function offset_lt(x::Endpoint, y::Endpoint)
   return (
       a.endpoint == b.endpoint && isleft(a) && isclosed(a) && !isleft(b) && isclosed(b) ||
       isless(a, b)
   )
end
```
This code is far easier to understand and is inline with the comment above. To do this however requires `unbunch` to take a `sortlt` keyword or another iteration where sorting is extracted from `unbunch`ing

src/interval.jl
```diff
@@ -469,6 +714,100 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
...
+all intervals in `y` that intersect with `x[i]`.
+
+"""
+function find_intersections(x_::AbstractIntervals, y_::AbstractIntervals)
+    xa = vcat(x_)
+    x = unbunch(xa; enumerate=true, sortby=offset)
+    y = unbunch(vcat(y_); enumerate=true, sortby=offset)
+    result = [Vector{Int}() for _ in 1:length(xa)]
+
+    find_intersections_helper(result, x, y)
```
omus, line 776:
```suggestion
    result = [Vector{Int}() for _ in 1:length(xa)]

    find_intersections_helper(result, x, y)
```
Function is mutating and should marked as such

test/sets.jl
```diff
@@ -0,0 +1,99 @@
...
+        # verify that all indices not returned in `find_intersections` correspond to
+        # sets in b that do not overlap with the given set in akk
+        @test all(ix -> isempty(intersect(a[ix[1]], b[Not(ix[2])])), 
+                  enumerate(intersections))
+    end
+
+    # verify empty interval set
+    @test isempty(union(Interval[]))
+    
+    translate(x, by) = Interval(first(x) + by, last(x) + by)
```
omus, line 37: FYI translation is handled by `+` and `-` currently:
```julia
julia> Interval(1, 5) + 2
Interval{Int64, Closed, Closed}(3, 7)
```

test/sets.jl
```diff
@@ -0,0 +1,99 @@
+using Intervals
+using InvertedIndices
+using Random
+
+@testset "Set operations" begin
+    area(x::Interval) = last(x) - first(x)
+    area(x::AbstractVector{<:AbstractInterval{T}}) where T = reduce(+, map(area, x), init = zero(T))
```
omus, line 7:
```suggestion
    area(x::AbstractVector{<:AbstractInterval{T}}) where T = mapreduce(area, +, x, init = zero(T))
```

omus (commented):
Haven't quite finished reviewing yet but I can already see this code needs additional iteration. In it's current state it'll be rather difficult to maintain. I'll do more review soon

src/interval.jl
haberdashPI: > Is that a use case you need for this though?

Yup, that's exactly the goal to have all end points in order (so that e.g. interval A's endpoints may surround interval B's endpoints).

haberdashPI: Though, actually, looking at how I ended up doing this, I think that suggestion would work... 🤔

src/interval.jl
haberdashPI: Odd, this I thought was just from the original branch, I will delete.

src/interval.jl
haberdashPI: Tried to improve the comments here to clarify

src/interval.jl
haberdashPI: Having worked through it a little more, I've simplified my approach but I still have a separate type of endpoint; I think this is the easiest to read / least friction. I wonder if it would be worth just briefly meeting and discussing this design choice.

The other option to simplify even more of this would be to remove the code that handles generic intervals, and just implement it for the Vector{<:Inteval{T,Open,Closed} and Vector{<:Inteval{T,Closed,Open} types. But that's exactly the case where the new `SimpleEndpoint` struct is useful, so it would not change that part.

src/interval.jl
```diff
@@ -36,6 +37,17 @@ julia> Dates.today() - Dates.Week(1) .. Dates.today()
...
+### Multi-interval set operations
+
+Set operations can also be performed over two pairs of interval arrays. These set operations
+take the form of `op(x::Vector{<:Interval}, y::Vector{<:Interval})` and intrepret `x` and 
+`y` as representing a set covered by the provided intervals. The return value is an array of
+non-overlaping intervals representing the result of `op` over these sets. You can also pass 
+a single interval to either arugment (e.g. op(x::Interval, y::Vector{<:Interval})).
+
+These multi-interval set operations currently only support intervals that are
+bounded.
```
omus, line 49: 92-char word wrap, spelling mistakes, quoting
```suggestion
Set operations can also be performed over two pairs of interval arrays. These set
operations take the form of `op(x::Vector{<:Interval}, y::Vector{<:Interval})` and
interpret `x` and `y` as representing a set covered by the provided intervals. The return
value is an array of non-overlapping intervals representing the result of `op` over these
sets. You can also pass a single interval to either argument
(e.g. `op(x::Interval, y::Vector{<:Interval})`).

These multi-interval set operations currently only support intervals that are bounded.
```

src/interval.jl
```diff
@@ -1,12 +1,13 @@
 """
     Interval{T, L <: Bound, R <: Bound}
 
-An `Interval` represents a non-iterable range or span of values (non-interable because,
-unlike a `StepRange`, no step is defined).
+An `Interval` represents a non-iterable range or span of values (non-interable
+because, unlike a `StepRange`, no step is defined).
```
omus, line 5:
```suggestion
An `Interval` represents a non-iterable range or span of values (non-interable because,
unlike a `StepRange`, no step is defined).
```

src/interval.jl
omus, line 10:
```suggestion
An `Interval` can be closed (both `first` and `last` are included in the interval), open
(neither `first` nor `last` are included), or half-open. This openness is defined by the
bounds information which is stored as the type parameters `L` and `R`.
```

test/sets.jl
```diff
@@ -0,0 +1,98 @@
...
+    testsets(intervals[1:25], intervals[26])
+
+    leftint(a,b) = Interval{Int, Closed, Open}(a, b)
+    leftint(a::Interval) = Interval{Int, Closed, Open}(first(a), last(a))
+    intervals = leftint.(starts, starts .+ rand(1:10_000, 25))
+    intervals = [intervals; leftint.(intervals .+ (round.(Int, area.(intervals) .*
+                                                (2.0.*rand(length(intervals)) .- 1.0))))]
+    testsets(intervals[1:25], intervals[26:end])
+    testsets(intervals[1], intervals[26:end])
+    testsets(intervals[1:25], intervals[26])
```
omus, line 78:
```suggestion
    a = leftint.(starts, starts .+ rand(1:10_000, 25))
    b = leftint.(a .+ round.(Int, area.(a) .* (2 .* rand(length(a)) .- 1)))
    testsets(a, b)
    testsets(a[1], b)
    testsets(a, b[1])
```

test/sets.jl
```diff
@@ -0,0 +1,98 @@
...
+
+    rightint(a,b) = Interval{Int, Open, Closed}(a, b)
+    rightint(a::Interval) = Interval{Int, Open, Closed}(first(a), last(a))
+    intervals = rightint.(starts, starts .+ rand(1:10_000, 25))
+    intervals = [intervals; rightint.(intervals .+ (round.(Int, area.(intervals) .*
+                                                (2.0.*rand(length(intervals)) .- 1.0))))]
+    testsets(intervals[1:25], intervals[26:end])
+    testsets(intervals[1], intervals[26:end])
+    testsets(intervals[1:25], intervals[26])
```
omus, line 87:
```suggestion
    a = rightint.(starts, starts .+ rand(1:10_000, 25))
    b = rightint.(a .+ round.(Int, area.(a) .* (2 .* rand(length(a)) .- 1))
    testsets(a, b)
    testsets(a[1], b)
    testsets(a, b[1])
```

test/sets.jl
```diff
@@ -0,0 +1,98 @@
...
+
+    intervals = leftint.(starts, starts .+ rand(1:10_000, 25))
+    intervals = [intervals; leftint.(intervals .+ (round.(Int, area.(intervals) .*
+                                                (2.0.*rand(length(intervals)) .- 1.0))))]
+    testsets(intervals[1:25], randint.(intervals[26:end]))
+    testsets(intervals[1], randint.(intervals[26:end]))
+    testsets(intervals[1:25], leftint(intervals[26]))
+    testsets(intervals[1:25], rightint.(intervals[26:end]))
+    testsets(intervals[1], rightint.(intervals[26:end]))
+    testsets(intervals[1:25], rightint(intervals[26]))
```
omus, line 97:
```suggestion
    a = leftint.(starts, starts .+ rand(1:10_000, 25))
    b = leftint.(a .+ round.(Int, area.(a) .* (2 .* rand(length(a)) .- 1)))
    testsets(a, randint.(b))
    testsets(a[1], randint.(b))
    testsets(a, leftint(b[1]))
    testsets(a, rightint.(b))
    testsets(a[1], rightint.(b))
    testsets(a, rightint(b[1]))
```

src/interval.jl
```diff
@@ -351,6 +363,266 @@ true
...
+    else
+        return [SimpleEndpoint(true, first(interval)), SimpleEndpoint(false, last(interval))]
+    end
+end
+
+function unbunch(interval::AbstractInterval, tracking::EndpointTracking; enumerate=false, lt=isless) where T
+    return unbunch_helper(interval, enumerate, 1, tracking)
+end
+function unbunch(intervals::AbstractIntervals, tracking::EndpointTracking; enumerate=false, lt=isless)
+    filtered = filter(i -> !isempty(intervals[i]), eachindex(intervals))
```
omus, line 461:
```suggestion
    filtered = filter(!isempty, intervals)
```

src/interval.jl
```diff
@@ -351,6 +363,266 @@ true
...
+    isempty(filtered) && return Union{}[]
+    result = mapreduce(vcat, filtered) do i
+        return unbunch_helper(intervals[i], enumerate, i, tracking)
+    end
```
omus, line 465: An option for removing the `enumerate` argument from `unbunch_helper`
```suggestion
    enumerator = enumerate ? tuple : (i, x) -> x
    result = mapreduce(vcat, filtered) do i
        return enumerator.(i, unbunch_helper(intervals[i], tracking))
    end
```
Requires changes to `unbunch_helper` to work

src/interval.jl
```diff
@@ -469,6 +752,86 @@ function Base.merge(a::AbstractInterval, b::AbstractInterval)
...
+Base.in(x, y::AbstractVector{<:AbstractInterval}) = any(yᵢ -> x ∈ yᵢ, y)
+function Base.issetequal(x::AbstractIntervals, y::AbstractIntervals)
+    x, y, tracking = unbunch(union(x), union(y))
+    return x == y || (all(isempty, bunch(x, tracking)) && all(isempty, bunch(y, tracking)))
+end
+Base.length(x::AbstractInterval) = 1
+
+# order edges so that closed boundaries are on the outside: e.g. [( )]
+intersection_order(x::Endpoint) = isleft(x) ? !isclosed(x) : isclosed(x)
+intersection_isless_fn(_::TrackStatically) = isless
```
omus, line 774:
```suggestion
intersection_isless_fn(::TrackStatically) = isless
```

haberdashPI:
Here's another round of changes! 

I ended up not removing the code to handle arbitrary endpoints. If you think I should really remove it, I can make that change. The infrastructure for `EndpointTracking` turned out to be super handy in resolving how exactly how to use `LeftEndpoint` and `RightEndpoint` rather than `SimpleEndpoint` and made determining return types easier (so at least some of the EndpointTracking is useful regardless).

haberdashPI:
@omus: I'm hoping to get this merged soon, any chance you can take a look at this in the coming week?

src/interval_sets.jl
```diff
@@ -0,0 +1,382 @@
...
+# order edges so that closed boundaries are on the outside: e.g. [( )]
+intersection_order(x::Endpoint) = isleft(x) ? !isclosed(x) : isclosed(x)
+intersection_isless_fn(::TrackStatically) = isless
+intersection_isless_fn(::TrackEachEndpoint) = function(x,y)
+    if isequal(x, y)
+        return isless(intersection_order(x), intersection_order(y))
+    else
+        return isless(x, y)
+    end
+end
```
omus, line 328: More inline with BlueStyle
```suggestion
function intersection_isless_fn(::TrackEachEndpoint)
    function (x,y)
        if isequal(x, y)
            return isless(intersection_order(x), intersection_order(y))
        else
            return isless(x, y)
        end
    end
end
```

test/runtests.jl
```diff
@@ -7,6 +7,7 @@ using Intervals: isfinite
 using Serialization: deserialize
 using Test
 using TimeZones
+TimeZones.build() # (requried in new versions of TimeZones)
```
omus, line 10: Not true. You could see random failures but this shouldn't be here
```suggestion
```

omus (changes requested):
Things are definitely in much better shape. I'm not quite done my review but I thought it best to post what I found

Project.toml
```diff
@@ -22,9 +22,13 @@ julia = "1"
...
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Serialization = "9e88b42a-f829-5b0c-bbe9-9e923198166b"
+StableRNGs = "860ef19b-820b-49d6-a774-d7a799459cd3"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 VisualRegressionTests = "34922c18-7c2a-561c-bac1-01e79b2c4c92"
 
 [targets]
-test = ["Documenter", "ImageMagick", "Infinity", "Plots", "Test", "VisualRegressionTests"]
+test = ["Documenter", "ImageMagick", "Infinity", "Plots", "Test", "VisualRegressionTests", "InvertedIndices", "Random", "Serialization", "StableRNGs"]
```
omus, line 34:
```suggestion
test = ["Documenter", "ImageMagick", "Infinity", "InvertedIndices", "Plots", "Random", "Serialization", "StableRNGs", "Test", "VisualRegressionTests"]
```

src/interval_sets.jl
```diff
@@ -0,0 +1,382 @@
+
+###### Set-related Helpers #####
+
+# most set-related methods can operate over both individual intervals and
+# vectors of intervals
+const AbstractIntervalsOf{T,L,R} = Union{AbstractInterval{T,L,R}, AbstractVector{<:AbstractInterval{T,L,R}}}
+const AbstractIntervals = Union{AbstractInterval, AbstractVector{<:AbstractInterval}}
```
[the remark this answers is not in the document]
haberdashPI: That is what I tried at first. The issue is that whenever you include parameters, they imply that the object is concrete. In most cases I watch a method that covers all objects of the given type, even when the inferred type is abstract. In a few cases (`endpoint_tracking`) I only want the concrete types.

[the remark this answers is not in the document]
haberdashPI: I've added a comment to explain this motivation.

src/interval_sets.jl
```diff
@@ -0,0 +1,382 @@
...
+    end
+
+    return bunch(result, endpoint_tracking)
+end
+# empty_interval: true if the given left and right endpoints would create an empty interval
+empty_interval(::SentinelEndpoint, _, _) = false # sentinal means there was no starting endpoint; there is thus no interval, and so no empty interval
+empty_interval(start, stop, ::TrackStatically) = start.endpoint == stop.endpoint
+function empty_interval(start, stop, ::TrackEachEndpoint)
+    return start.endpoint == stop.endpoint && any(!isclosed, (start, stop))
+end
```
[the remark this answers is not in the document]
haberdashPI: > It's unclear to me if start could be a RightEndpoint which may require more changes

Nope, it's always a LeftEndpoint. I've re-arranged code, adjusted comments to make that point more obvious (I think it improves overall clarity).

src/interval_sets.jl
```diff
@@ -0,0 +1,382 @@
...
+# set operations over multi-interval sets
+Base.intersect(x::AbstractIntervals, y::AbstractIntervals) = mergesets((inx, iny) -> inx && iny, x, y)
+Base.union(x::AbstractIntervals, y::AbstractIntervals) = mergesets((inx, iny) -> inx || iny, x, y)
+Base.setdiff(x::AbstractIntervals, y::AbstractIntervals) = mergesets((inx, iny) -> inx && !iny, x, y)
+Base.symdiff(x::AbstractIntervals, y::AbstractIntervals) = mergesets((inx, iny) -> inx ⊻ iny, x, y)
+Base.issubset(x::AbstractIntervals, y::AbstractIntervals) = isempty(setdiff(x, y))
+# may or may not be from Base (see top of `Intervals.jl`)
+Base.isdisjoint(x::AbstractIntervals, y::AbstractIntervals) = isempty(intersect(x, y))
+
+Base.in(x, y::AbstractVector{<:AbstractInterval}) = any(yᵢ -> x ∈ yᵢ, y)
```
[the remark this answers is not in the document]
haberdashPI: I think the difference here is that the actual representation of a set is a vector of intervals. (That's how all the set operation functions work). 

I don't actually need this currently (and would be easy to define some helper, `in_interval`) so I'm not strongly opinionated about whether this gets included, but I think it's coherent with how interval arrays are treated elsewhere.

src/interval_sets.jl
```diff
@@ -0,0 +1,382 @@
...
+function Base.issetequal(x::AbstractIntervals, y::AbstractIntervals)
+    x, y, tracking = unbunch(union(x), union(y))
+    return x == y || (all(isempty, bunch(x, tracking)) && all(isempty, bunch(y, tracking)))
+end
+Base.length(x::AbstractInterval) = 1
```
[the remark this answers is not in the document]
haberdashPI: I use it in a few placed. I can jus rename this, since I'm not sure it's valuable external to the package.

haberdashPI:
Thanks for the feedback! I'll see how CI does with those changes and then fix any issues it finds.

The one unresolved point is about `Base.in`. I'm not strongly for it, but what do you think of my point about it being consistent with the other methods? If you still think it should be removed, I'll delete it; otherwise I can document the behavior.

src/interval_sets.jl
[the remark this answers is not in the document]
omus: The `in` function should be used to check if an element is part of a collection. I'd expect the following behaviour:
```julia
julia> 1..2 in [1..2]
true

julia> 1..2 in [0..3]
false

julia> 1 in [0..3]
false

julia> 1 in 0..3
true
```
I think the `in_interval` function you want is just `issubset` which seems sensible as you can treat an interval itself as a collection

src/interval_sets.jl
```diff
@@ -0,0 +1,425 @@
...
+    return union!(convert(Vector{AbstractInterval}, intervals))
+end
+# allow a concretely typed array for `Interval` objects (as opposed to e.g. anchored intervals
+# which may change type during the union process)
+function Base.union(intervals::AbstractVector{T}) where T <: Interval
+    input = convert(Vector{T}, intervals)
+    if input === intervals
+        input = copy(intervals)
+    end
+    return union!(input)
```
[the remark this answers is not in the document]
haberdashPI: The desired result is that 
  1. The input to `union!` has the right eltype (hence the convert).
  2. The input to `union!` is a copy.

`convert` may or may not produce a copy: e.g. `convert(Vector{Float64}, Float64[1,2,3])` doesn't produce a copy. 
`input === intervals` checks whether `convert` made a copy and makes one if it hasn't been made.

omus:
> The one [unresolved point is about `Base.in`](https://github.com/invenia/Intervals.jl/pull/179#discussion_r865029455). I'm not strongly for it, but what do you think of my point about it being consistent with the other methods? If you still think it should be removed, I'll delete it; otherwise I can document the behavior.

We've resolve this. The end result is to use `issubset` for this behaviour.

I spent a little time with this PR over the weekend and came to a conclusions:

1. The tests added for sets here do not cover the common corner cases for intervals especially with bounds. Extending the comparison tests should be most of what we need to do.
2. Set methods returning non-booleans should all return vectors of intervals. We need to gracefully deprecate the current behaviour (#189) and have this PR not make things worse

Further detail on point 1:

I've implemented the comparison tests in https://github.com/invenia/Intervals.jl/pull/190. Doing so has revealed that the set code implemented here doesn't work correctly when dealing with interval endpoints using the same values. One such example:

```julia
julia> union([Interval{Closed,Closed}(1, 3)], [Interval{Open, Open}(3, 5)])
2-element Vector{Interval{Int64}}:
 Interval{Int64, Closed, Closed}(1, 3)
 Interval{Int64, Open, Open}(3, 5)
```

Other failures can be found by looking for `@test_broken` in `tests/comparisons.jl`

Some more detail on point 2:

The original set methods added to Intervals.jl were treating interval instances as sets themselves. Specifically the `intersect(::Interval, ::Interval)` method would return an interval. In this case the logic is sound but the other set methods can return multiple intervals under certain conditions. Because the set functions should be consistent in their return arguments advocate for the following:

- All existing set functions returning a single interval should be deprecated
- All new set functions returning a vector of intervals should only accept arguments of vectors
- When the deprecations are dropped we can introduce set functions `f(::Interval, ::Interval) -> Vector{Interval}`

@haberdashPI to wrap up this PR we need to:

- [ ] Get the set function working with touching intervals (point 1)
- [ ] Review the changes in #190 to see if they are appropriate. We can apply the new changes to this PR via a cherry-pick

haberdashPI:
Placing these here, since I can't edit your comments:

- [ ] Get the set function working with touching intervals (point 1)
- [x]  Review the changes in https://github.com/invenia/Intervals.jl/pull/190 to see if they are appropriate. We can apply the new changes to this PR via a cherry-pick

haberdashPI:
Thinking a little more deeply on these corner cases, I have a bit of a philosophical conundrum:

If we treat the returned value as a representation of the set of all points that would be included in the implemented set operation, there are actually an infinite number of such representations. That is, in pseudo code we have

```julia
issetequal({[0,1), [1, 2]}, {[0, 2]})
```
(I'm using `{}` to represent arrays above, and `[` and `(` to represent the open/closed distinction)

You can divide any interval into an array of as many subintervals and they remain setequal, according to this representation. The question then becomes, which setequal value is most reasonable? An obvious choice would be the smallest such representation.

However, your tests assume that `union([0,1), [2, 3])` returns `[0, 3]`, while `symdiff([0,1), [2,3])` returns `{[0, 1), [2,3]}`, even though these two results are `setequal`. Arguably the right answer in both cases is `[0, 3]`

I *think* your test is the expected/more useful behavior. But it is a somewhat conceptually odd choice...

I could just special case this one situation, but I think it gets weirder when you start thinking about multiple intervals as the input: when should you and when should you not union two abutting intervals across different multi-interval set operations?

I'm currently trying to think through some sort of systematic framing for why the cardinality of the output would differ between these two cases.

haberdashPI:
Thinking through this more: 

I think it is reasonable to be replace testing for `==`for the return value of symdiff with the following

```julia
setequal(x, y) && legnth(x) <= length(y)
```

Because the two outputs symdiff could return (two vs. one interval representation) are setequal, they will produce the same outcome for many computations (just with a smaller N). Any operation that wouldn't produce the same result is somehow relying on the cardinality of the returned value from a set operation. That sounds pretty fragile to me, since in general one can't rely on a particular cardinality of the returned value.

haberdashPI:
I think this test on line 594 of comparisions.jl is not quite right:

```julia
            expected_xor = [Interval{Closed, Open}(first(a), first(a))]
```

That interval `isempty` but, e.g. `!isempty(setdiff([1, 5) , (1, 5)))` (which is right, since the singular value at 1 should be included).

haberdashPI:
>> However, your tests assume that `union([0,1), [2, 3])` returns `[0, 3]`
>
> Which tests assume this? This would definitely be incorrect

My bad: I meant something different

>  However, your tests assume that `union([0,1), [1, 2])` returns `[0, 2]`, while `symdiff([0,1), [1, 2])` returns `{[0, 1), [1,2]}`, even though these two results are `setequal`. Arguably the right answer in both cases is `[0, 2]`

test/comparisons.jl
```diff
@@ -419,7 +421,8 @@ end
                 @test setdiff([earlier], [later]) == expected_xor[1:1]
                 @test setdiff([later], [earlier]) == expected_xor[2:2]
 
-                @test_broken symdiff([earlier], [later]) == expected_xor
+                @test issetequal(symdiff([earlier], [later]), expected_xor)
+                @test length(symdiff([earlier], [later])) <= length(expected_xor)
```
omus, line 425:
```suggestion
                @test symdiff([earlier], [later]) == expected_xor
```

test/comparisons.jl
```diff
@@ -629,14 +632,18 @@ end
...
                 @test intersect([a], [b]) == [expected_overlap]
 
-                @test_broken setdiff([a], [b]) == expected_xor[1:1]
+                @test issetequal(setdiff([a], [b]), expected_xor[1:1])
+                @test length(setdiff([a], [b])) <= length(expected_xor[1:1])
                 @test setdiff([b], [a]) == []
 
-                @test_broken symdiff([a], [b]) == expected_xor
+                @test issetequal(symdiff([a], [b]), expected_xor)
+                @test length(symdiff([a], [b])) <= length(expected_xor)
```
omus, line 646:
```suggestion
                @test symdiff([a], [b]) == expected_xor
```

test/comparisons.jl
```diff
@@ -696,14 +703,18 @@ end
...
             # TODO: These functions should be compatible with unbounded intervals
+            # TODO: will have to think carefully about the `expected_` variables
+            # when we allow for unbounded values
             if isbounded(a) && isbounded(b)
                 @test union([a], [b]) == [expected_superset]
                 @test intersect([a], [b]) == [expected_overlap]
 
-                @test_broken setdiff([a], [b]) == expected_xor[1:1]
+                @test issetequal(setdiff([a], [b]), expected_xor[1:1])
+                @test length(setdiff([a], [b])) <= length(expected_xor[1:1])
```
omus, line 713:
```suggestion
                @test setdiff([a], [b]) == expected_xor[1:1]
```

test/comparisons.jl
```diff
@@ -696,14 +703,18 @@ end
...
                 @test setdiff([b], [a]) == []
 
-                @test_broken symdiff([a], [b]) == expected_xor
+                @test issetequal(symdiff([a], [b]), expected_xor)
+                @test length(symdiff([a], [b])) <= length(expected_xor)
```
omus, line 717:
```suggestion
                @test symdiff([a], [b]) == expected_xor
```

test/comparisons.jl
```diff
@@ -767,14 +778,18 @@ end
...
             # TODO: These functions should be compatible with unbounded intervals
+            # TODO: will have to think carefully about the `expected_` variables
+            # when we allow for unbounded values
             if isbounded(a) && isbounded(b)
                 @test union([a], [b]) == [expected_superset]
                 @test intersect([a], [b]) == [expected_overlap]
 
-                @test_broken setdiff([a], [b]) == expected_xor[1:1]
+                @test issetequal(setdiff([a], [b]), expected_xor)
+                @test length(setdiff([a], [b])) <= length(expected_xor)
```
omus, line 788:
```suggestion
                @test setdiff([a], [b]) == expected_xor
```

test/comparisons.jl
```diff
@@ -767,14 +778,18 @@ end
...
                 @test setdiff([b], [a]) == []
 
-                @test_broken symdiff([a], [b]) == expected_xor
+                @test issetequal(symdiff([a], [b]), expected_xor)
+                @test length(symdiff([a], [b])) <= length(expected_xor)
```
omus, line 792:
```suggestion
                @test symdiff([a], [b]) == expected_xor
```

test/comparisons.jl
```diff
@@ -841,9 +856,11 @@ end
                 @test intersect([a], [b]) == [expected_overlap]
 
                 @test setdiff([a], [b]) == []
-                @test_broken setdiff([b], [a]) == expected_xor[1:1]
+                @test issetequal(setdiff([b], [a]), expected_xor[1:1])
+                @test length(setdiff([b], [a])) <= length(expected_xor[1:1])
```
omus, line 860:
```suggestion
                @test setdiff([b], [a]) == expected_xor[1:1]
```

test/comparisons.jl
```diff
@@ -841,9 +856,11 @@ end
...
 
-                @test_broken symdiff([a], [b]) == expected_xor
+                @test issetequal(symdiff([a], [b]), expected_xor)
+                @test length(symdiff([a], [b])) <= length(expected_xor)
```
omus, line 863:
```suggestion
                @test symdiff([a], [b]) == expected_xor
```

test/comparisons.jl
```diff
@@ -910,9 +927,11 @@ end
                 @test union([a], [b]) == [expected_superset]
 
                 @test setdiff([a], [b]) == []
-                @test_broken setdiff([b], [a]) == expected_xor[1:1]
+                @test issetequal(setdiff([b], [a]), expected_xor[1:1])
+                @test length(setdiff([b], [a])) == length(expected_xor[1:1])
```
omus, line 931:
```suggestion
                @test setdiff([b], [a]) == expected_xor[1:1]
```

test/comparisons.jl
```diff
@@ -910,9 +927,11 @@ end
...
 
-                @test_broken symdiff([a], [b]) == expected_xor
+                @test issetequal(symdiff([a], [b]), expected_xor)
+                @test length(symdiff([a], [b])) == length(expected_xor)
```
omus, line 934:
```suggestion
                @test symdiff([a], [b]) == expected_xor
```

haberdashPI:
Alright, placed some unions on the appropriate expected_xor's and that seems to have cleared up the remaining failed tests.

omus:
I'll wait until Monday to squash and merge this PR. This should allow anyone who was waiting for the dust to settle on this PR the chance to done a final review before we merge this and release version 1.7.

rofinn:
NOTE: This release actively breaks our code base. Since this PR actively changes the behaviour of methods operating on existing types (ie: vector of intervals) I'm rolling this change back and it should be made as a breaking change.

haberdashPI:
Hi @rofinn,

First off, sounds like this was pretty disruptive; so sorry for that. A few thoughts on how to move forward:

> but since we don't distinguish an IntervalSet from a Vector{<:Interval} there is no way for us to distinguish this ambiguous case

I think this is not quite right. You can pass the objects as `Set`s to get the old behavior: e.g.

```julia
julia> intersect(Set([1..2, 2..3, 3..4, 4..5]), Set([2..3, 3..4]))
Set{Interval{Int64, Closed, Closed}} with 2 elements:
  Interval{Int64, Closed, Closed}(2, 3)
  Interval{Int64, Closed, Closed}(3, 4)
```

So perhaps what is needed here is a deprecation like this?

```julia
@deprecate intersect(a::AbstractArray, b::AbstractArray) intersect(Set(a), Set(b))
```

I guess a potential issue there would be that there might be a performance hit for first converting to `Set` objects?

In the meantime, perhaps the new behavior can be made available by creating a simple version of the `IntervalSet` objects and dispatching `Base.intersect` (and friends) on those.

rofinn:
Yeah, I think at minimum we should either define a minimal `IntervalSet` or just use the package namespace to distinguish the behaviour. We can always talk about adding a deprecation in a future minor release, but would need to leave making this the default behaviour with the base function until a major release. I'll revert this MR and tag to minimize impact to the ecosystem. I'll then start a new MR with your changes where we can flush out how to make it non-breaking.

omus:
> @omus If you're okay with that I'll plan to just yank this release from the registry while try to move these changes into a 2.0 release.

Very reasonable. I apologize for missing this breaking behaviour change. Sounds like we have some additional tests that need to be added to the Intervals.jl test to ensure we don't break expected this behaviour again. Is there a better way to communicate when large changes like this are going to be released? Seems like the typical people-who-care-are-watching isn't working

omus:
> Yeah, I think at minimum we should either define a minimal `IntervalSet` or just use the package namespace to distinguish the behaviour.

I'm in favour of making a minimal `IntervalSet` type

rofinn:
> would be good to add some tests in here for your use case as well

Yeah, this is a bit of a weird edge case where we'd need to test for a behaviour that wasn't defined here. I've become rather skittish of extending functions if there's concern that the new method doesn't align with the original concept/definition in Base.

> Seems like the typical people-who-care-are-watching isn't working

Yeah, I don't think anyone was ever assigned to keep an eye on this repo after you left. I guess for now it'd probably be best to just assign myself or @fchorney to larger reviews for now as we're the next biggest contributors. I'll raise this maintenance concern more broadly to see if there's a more sustainable option.

> I'm in favour of making a minimal IntervalSet type

I was hoping that namespacing would be the simpler solution, but given how many things we already extend in base I think you're probably correct. Moving things from the Base to Intervals namespaces would likely be more breaking.

PR #186, opened 2022-01-20
Set minimum supported julia version to 1.6 (LTS)

pabloferz:


.github/workflows/CI.yml
```diff
@@ -30,11 +30,6 @@ jobs:
             arch: x86
           - os: windows-latest
             arch: x86
-        include:
-          # Add a 1.5 job because that's what Invenia actually uses
```
pabloferz, line 34: Is this still true?

omus (commented):
Changes look good. The 32-bit failures appear unrelated but should be sorted out

omus:
🎉  Very nice work! I suggest we should also increase the minimum version of Julia in the Project.toml to be `1.6` as we no longer support earlier versions and don't want people accidentally installing this on older Julia revisions. Besides that I'm happy to approve.

Hopefully, someone from Invenia can validate the unresolved discussion.

omus (approved):

omus:
I'll merge this by EOD unless there are objections. If Invenia wants to keep testing against Julia 1.5 we can always add that back in.

PR #187, opened 2022-05-03
Drop unnecessary code paths to support Julia < 1.6

omus:
Things that should have been included in #186

### DIFF
--- a/src/Intervals.jl
+++ b/src/Intervals.jl
@@ -57,7 +57,6 @@ export Bound,
        union!,
        less_than_disjoint,
        greater_than_disjoint,
-       isdisjoint,
        superset,
        .., ≪, ≫, ⊆, ⊇, ⊈, ⊉
 end
--- a/test/test_utils.jl
+++ b/test/test_utils.jl
@@ -2,14 +2,8 @@
 # format is the same as the last version of Julia we've tested against.
 const SERIALIZED_HEADER = if VERSION >= v"1.5"
     "7JL\n\x04\0\0"
-elseif VERSION >= v"1.4"
-    "7JL\t\x04\0\0"
-elseif VERSION >= v"1.2"
-    "7JL\b\x04\0\0"
-elseif VERSION >= v"1.0"
-    "7JL\a\x04\0\0"
 else
-    error("Julia versions earlier than 1.0 are unsupported")
+    error("Julia versions earlier than 1.5 are unsupported")
 end
 
 # Declare a new `isnan` and `isinf` functions to avoid type piracy. These new function work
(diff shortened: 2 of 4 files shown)


test/test_utils.jl
```diff
@@ -2,14 +2,8 @@
 # format is the same as the last version of Julia we've tested against.
 const SERIALIZED_HEADER = if VERSION >= v"1.5"
```
mjram0s, line 3:
```suggestion
const SERIALIZED_HEADER = if VERSION >= v"1.6"
```
The linked MR changed min supported version to `1.6`

test/test_utils.jl
```diff
@@ -2,14 +2,8 @@
...
     "7JL\n\x04\0\0"
-elseif VERSION >= v"1.4"
-    "7JL\t\x04\0\0"
-elseif VERSION >= v"1.2"
-    "7JL\b\x04\0\0"
-elseif VERSION >= v"1.0"
-    "7JL\a\x04\0\0"
 else
-    error("Julia versions earlier than 1.0 are unsupported")
+    error("Julia versions earlier than 1.5 are unsupported")
```
mjram0s, line 6:
```suggestion
    error("Julia versions earlier than 1.6 are unsupported")
```

src/Intervals.jl
```diff
@@ -57,7 +57,6 @@ export Bound,
        union!,
        less_than_disjoint,
        greater_than_disjoint,
-       isdisjoint,
```
fchorney, line 60: was this just left over from some other change previously?

fchorney (approved):
With MJs suggestions, this LGTM

src/Intervals.jl
iamed2: This was added to Base in Julia 1.5, so is now always available and doesn't need to be exported
https://docs.julialang.org/en/v1/base/collections/#Base.isdisjoint

test/test_utils.jl
omus: I did noticed this and kept it as 1.5 as the header happens to be the same. I thought since we already know it's 1.5 compatible so I kept it that way. I can definitely change it just let me know what you prefer.

src/Intervals.jl
omus: Should have been removed as part of: https://github.com/invenia/Intervals.jl/pull/186

test/test_utils.jl
iamed2: Currently this version will never run on <1.6 due to the restriction in Project.toml. Is it worth even having this check anymore?

omus: > Currently this version will never run on <1.6 due to the restriction in Project.toml. Is it worth even having this check anymore?

That's a good point I didn't consider. Probably in some future version of Julia this header will change again but as of now it hasn't change as of Julia 1.5 so we can drop the logic until it's needed again. I'll leave an appropriate comment

iamed2 (approved):

PR #190, opened 2022-05-16
Multi-interval set operators (extended)

omus:
Updates some of the work done in #179. As that PR is a fork unfortunately I can't use that PR as a base


omus:
Changes here were incorporated into #179

PR #191, opened 2022-06-01
Revert "Multi-interval set operators"

rofinn:
Reverts invenia/Intervals.jl#179


morris25 (approved):
Relevant comment for reversion https://github.com/invenia/Intervals.jl/pull/179#issuecomment-1143001180

rofinn:
That's what I thought, but @oxinabox said that yanking releases is typically only done for security or compat issues.

omus:
> That's what I thought, but @oxinabox said that yanking releases is typically only done for security or compat issues. Honestly, yanking releases appears to be an undocumented black art that I'm happy to avoid :)

Definitely could be documented better. As the 1.7.0 release is still available it could still be used and break our semver requirements. I can yank 1.7.0 if that sounds good to you. In the future I suggest using yanking for this kind of rollback.

oxinabox:
>  As the 1.7.0 release is still available it could still be used and break our semver requirements. 

No more or less than any other release that is not the final patch release.
It's fine to yank it, but that exact same logic can be used to yank every patch release but the last.
Which, I have argued, should actually be the behavior of Pkg by default.
Namely that, post v1.0.0, only the last patch release should in installable unless some override flag is passed.

PR #192, opened 2022-06-01
Revert "Revert "Multi-interval set operators""

rofinn:
It's ugly, but necessary. We're reverting #191 which in turn reverted #179.

We just need to address the https://github.com/invenia/Intervals.jl/pull/179#issuecomment-1143001180 from the original PR and double check that no other cases exists (ie: review all new Base method overloads).

NOTE: Something else that we should probably change is the `union` behaviour as our definition of `union` doesn't really match the `Base` definition.


haberdashPI:
Moving my note over:

This should also add test cases for treating each interval as an element in a set (represented by an array).

> Something else that we should probably change is the union behaviour as our definition of union doesn't really match the Base definition.

Can you elaborate on this point?

haberdashPI:
Also, happy to start making some changes here with proposed fixes. Does that work? Or did you have plans to work on this branch?

rofinn:
> Can you elaborate on this point?

Yeah, the main difference is that the `Base.union` is combining all unique elements of the inputs to a single collection, but the elements don't change. In our case, we're basically applying a reduction on a single collection of intervals. I don't think the term `union` is wrong in this case, but the behaviour is very different from the base definition, hence I think it should probably also stay within the Intervals namespace.

omus:
> I was planning on making the minimal changes required to make this non-breaking and then I was gonna ask you and @omus for a review.

Can I ask that you cherry-pick the commit from #179 and then apply your changes as a separate commit? Will be much easier to review this by seeing just the differences between that PR and this.

rofinn:
I'd rather not mess with the history at this point. My plan was to just add an extra commit that you could review independently from all the rest.

haberdashPI:
Best of both worlds? @rofinn if you create a separate PR that merges into this one, it would be easy to review the changes, and then we can merge into this branch, and from there to main. ?? Just a suggestion. Can also just review locally and pick the right commits for the diff.

rofinn:
History is a bit ugly, but this should be good to go now.

PR #193, opened 2022-06-02
Minimal IntervalSet type

rofinn:
Currently added a simple IntervalSet type and most tests pass now. I've included example tests with both `Vector{<:AbstractInterval}` and `IntervalSet`s to identify where they differ as `union` still works with both.


rofinn:
GitHub actions seems to be borked.

src/interval_sets.jl
```diff
@@ -1,7 +1,22 @@
 
 ###### Set-related Helpers #####
 
-const IntervalSet = AbstractVector{<:AbstractInterval}
+struct IntervalSet{T<:AbstractInterval}
+    items::Vector{<:AbstractInterval}
+end
```
omus, line 6:
```suggestion
struct IntervalSet{T<:AbstractInterval} <: AbstractSet{T}
    items::Vector{T}
end
```

src/interval_sets.jl
```diff
@@ -1,7 +1,22 @@
...
+
+IntervalSet(v::AbstractVector) = IntervalSet{eltype(v)}(v)
+IntervalSet(interval::T) where T <: AbstractInterval = IntervalSet{T}([interval])
+IntervalSet(interval::IntervalSet) = interval
+IntervalSet(itr) = IntervalSet{eltype(itr)}(collect(itr))
```
omus, line 11:
```suggestion
IntervalSet(itr) = IntervalSet{eltype(itr)}(collect(itr))
IntervalSet() = IntervalSet{AbstractInterval}(AbstractInterval[])
```

test/comparisons.jl
```diff
@@ -410,16 +443,24 @@ end
 
             # Using a vector of intervals as sets
             @test union([earlier, later]) == [expected_superset]
+            @test union(IntervalSet([earlier, later])) == IntervalSet(expected_superset)
 
             # TODO: These functions should be compatible with unbounded intervals
             if isbounded(earlier) && isbounded(later)
-                @test union([earlier], [later]) == [expected_superset]
-                @test intersect([earlier], [later]) == [expected_overlap]
+                @test union([earlier], [later]) != [expected_superset]
```
omus, line 450: Should be updated with what is expected

omus (commented):
First pass review

haberdashPI (commented):
> > would be good to add some tests in here for your use case as well
>
> Yeah, this is a bit of a weird edge case where we'd need to test for a behaviour that wasn't defined here. I've become rather skittish of extending functions if there's concern that the new method doesn't align with the original concept/definition in Base.

I'm in agreement with @omus on this: I do see why this behavior is not something we would necessarily want to *keep* (it isn't really by design and I certainly didn't predict this use), and at some point this behavior should probably be deprecated. But it is the whole reason we have a breaking change that requires a revert of 1.7.0. Might as well be explicit about that in the tests, and then other folks can understand why things are set up this way for now. These tests can be marked as broken / deprecated when the change is actually made.

src/interval_sets.jl
rofinn: I'm not sure what the benefit of subtyping `AbstractSet` is. I feel like these operations are sufficiently different that it might be better to treat it as a set-like object?

src/interval_sets.jl
rofinn: Sure. I also wasn't sure if `getindex` would also be worth including.

test/comparisons.jl
rofinn: So we're just gonna update all these tests then? I'll note that these tests are extremely tedious and verbose.

rofinn:
Thanks for the review! Folks are welcome to add commits to this branch while I tests these changes within our ecosystem to make sure we aren't missing any other breaking changes.

src/interval_sets.jl
```diff
@@ -1,7 +1,22 @@
...
+
+Base.copy(intervals::IntervalSet{T}) where T = IntervalSet{T}(copy(intervals.items))
+Base.length(intervals::IntervalSet) = length(intervals.items)
+Base.iterate(intervals::IntervalSet, args...) = iterate(intervals.items, args...)
+Base.eltype(::IntervalSet{T}) where T = T
+Base.:(==)(a::IntervalSet, b::IntervalSet) = a.items == b.items
```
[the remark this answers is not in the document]
omus: I think this should happen on construction just like how `Set` works:
```julia
julia> Set([1,1,2])
Set{Int64} with 2 elements:
  2
  1
```

src/interval_sets.jl
omus: As `Set`s don't define `getindex` so I'm not sure why we would define that

src/interval_sets.jl
omus: > I feel like these operations are sufficiently different that it might be better to treat it as a set-like object?

I'd like to dig into this further. In my mind an `IntervalSet` is just a specialize form of a `Set` that understands that an one interval can overlap with another and takes advantage of that property by merging those intervals into one.

omus: We can leave making this a subtype of `AbstractSet` for a future PR.

One thought that occurred to me is I'm not sure what your goals for a minimal `IntervalSet` are. My assumption was that `IntervalSet` would gate the new breaking behaviour behind this new type. Possibly you had something different in mind?

omus:
> Thanks for the review! Folks are welcome to add commits to this branch while I tests these changes within our ecosystem to make sure we aren't missing any other breaking changes.

Sounds good. I suspect I'll be making some additions next week

src/interval_sets.jl
rofinn: No, that was my primary goal as well. I'd just like to make as few assumption about it for now. My concern is just that the merging behaviour doesn't really fit any existing set types in the ecosystem is all.

haberdashPI:
I'd be happy to add some more tests to this branch: though I don't think I have the appropriate permissions to push here.

rofinn:
Good news! Looks like these changes are no longer breaking for us, at least not in our production codebase.

src/interval_sets.jl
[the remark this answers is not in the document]
rofinn: > We should perform union on the vector passed into IntervalSet upon construction to maintain ordering and combine overlapping intervals

That'll break the current behaviour for `union!` because it'll require making an extra copy.

test/comparisons.jl
rofinn: Okay, I've included both the previous expected values (which we reverted) and the current behaviour (ie: base methods) on one line to make reviewing a bit easier.

src/interval_sets.jl
```diff
@@ -1,7 +1,28 @@
 
 ###### Set-related Helpers #####
+"""
+    IntervalSet{T<:AbstractInterval} <: AbstractSet{T}
+
+A set type for performing multi-interval specific operations.
+https://en.wikipedia.org/wiki/Interval_arithmetic#Interval_operators
+"""
+struct IntervalSet{T<:AbstractInterval} <: AbstractSet{T}
+    # TODO: Use Dict internally once union!(Vector{AbstractInterval}) is fully deprecated
```
rofinn, line 10: In relation to a point made by @omus, we need to wrap a vector without mutating or copying in order to properly deprecated the old `union!` behaviour. Once that's gone we should be good to use a more appropriate internal type like a `Dict`.

src/interval_sets.jl
```diff
@@ -264,35 +295,37 @@ right_endpoint(t, ::TrackRightOpen{T}) where T = RightEndpoint{T,Open}(endpoint(
...
-# which may change type during the union process)
-function Base.union(intervals::AbstractVector{T}) where T <: Interval
-    return union!(copy(intervals))
+Base.union(intervals::IntervalSet{<:Interval}) = union!(copy(intervals))
+
+# In the case where we're dealing with a non-concrete interval type like AnchoredIntervals then simply
+# allocate a AbstractInterval vector
+function Base.union(intervals::IntervalSet{<:AbstractInterval})
+    T = AbstractInterval
+    dest = Vector{T}(undef, length(intervals.items))
```
rofinn, line 309: We need to explicitly construct the destination otherwise `convert` will reuse source array. This was causing really weird test errors in which `union(expected_xor)` was mutating the input during a test, but only for certain types. I'm a little surprised the tests didn't previous catch this.

test/comparisons.jl
```diff
@@ -247,20 +263,30 @@ end
 
             # Using a vector of intervals as sets
             @test union([earlier, later]) == [expected_superset]
+            @test union(IntervalSet([earlier, later])) == IntervalSet([expected_superset])
 
             # TODO: These functions should be compatible with unbounded intervals
             if isbounded(earlier) && isbounded(later)
-                @test union([earlier], [later]) == [expected_superset]
+                @test union([earlier], [later]) == expected_xor != [expected_superset]
```
rofinn, line 270: For cases like this where something was previously `==` I've included both the `!=` case and what's actually returned. I'm not sure if we should keep these tests in since this is behaviour that was reverted?

rofinn (commented):
I'll leave this up to other folks to decide if they want add more tests.

src/interval_sets.jl
iamed2: I think the problem here is that an `AbstractSet` is a finite set of items and an `Interval` is (theoretically) an infinite set, and an `IntervalSet` represents the union of `Interval`s and is thus also an infinite set. You can iterate an `AbstractSet` but not an `Interval` (side note: I do think we shouldn't implement iteration for `IntervalSet` and should instead pass it off to a function; I would think `IntervalSet` would act more like an `Interval` than it would a collection of `Interval`s.

haberdashPI:
I've added some additional proposed changes here:

https://github.com/invenia/Intervals.jl/pull/196

haberdashPI:
I made some further changes over in #196, because the internal function `unbunch` needs to support vectors as well as sets: `find_intersections` should not accept `IntervalSet` objects, because the input is ordered, and no intervals should ever be merged, regardless of future optimizations of this kind to `IntervalSet`.

src/interval_sets.jl
rofinn: As per @iamed2's suggestion, I've made `IntervalSet`s non-iterable for now. It's a bit uglier in terms of the internals and tests, but it seems more correct for now. We can also make it iterable again later if this is too annoying.

src/interval_sets.jl
```diff
@@ -427,16 +434,17 @@ Base.intersect(x::IntervalSet, y::IntervalSet) = mergesets((inx, iny) -> inx &&
 Base.union(x::IntervalSet, y::IntervalSet) = mergesets((inx, iny) -> inx || iny, x, y)
 Base.setdiff(x::IntervalSet, y::IntervalSet) = mergesets((inx, iny) -> inx && !iny, x, y)
 Base.symdiff(x::IntervalSet, y::IntervalSet) = mergesets((inx, iny) -> inx ⊻ iny, x, y)
+Base.issubset(x::T, y::IntervalSet{<:AbstractInterval{T}}) where {T} = any(i -> x in i, y.items)
```
haberdashPI, line 437: Doesn't this mean that if x doesn't have the same concrete type as the element type of y it will throw a method error? I think `issubset(2, [1.0..3.0, 4.0..5.0])` should be `true` not a method error.

rofinn: Done. One thing I also noticed is that we aren't testing things like `issubset(interval, intervals)` and `issubset(intervals, interval)` which would fallback to calling the base `setdiff` and erroring. I'll go through and add more tests for the other set operations in case there are other incorrect fallback conditions.

haberdashPI (approved):
LGTM!

rofinn:
@omus Are you good with these changes?

src/interval_sets.jl
```diff
@@ -1,7 +1,98 @@
 
 ###### Set-related Helpers #####
+"""
+    IntervalSet{T<:AbstractInterval}
+
+An set of points represented by a sequence of intervals. Set operations over interval sets
```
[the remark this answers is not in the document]
rofinn: I think that's for markdown pages like /docs and our wiki. In either case, this package predates most of our conventions.

src/interval_sets.jl
```diff
@@ -1,7 +1,98 @@
...
+function Base.show(io::Base.AbstractPipe, ::MIME"text/plain", x::IntervalSet)
+    intervals = union(x)
+    n = length(intervals.items)
+    iocompact = IOContext(io, :compact => true)
+    print(io, "$n-interval ")
+    show(io, MIME"text/plain"(), typeof(x))
+    println(io, ":")
+    nrows = displaysize(io)[1]
+    half = fld(nrows, 2) - 2
+    if nrows ≥ n && half > 1
```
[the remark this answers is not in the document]
rofinn: I think `≥` is pretty common in this package as it uses a lot of unicode characters.

https://github.com/invenia/Intervals.jl/blob/master/src/interval.jl#L76

mattBrzezinski (approved):

rofinn:
Alright, I think this has been through a pretty thorough review now. I'm gonna merge and tag this, so folks can start using the new type and give feedback. I think the separate type and explicit deprecations should make it pretty unlikely to be breaking.

Issue #175, opened 2021-08-11
Add `AnchoredInterval` parsing


rofinn:
In some cases, if you know that your intervals are all `HE` or `HB` we could improve the parsing performance and memory requirements by parsing our intervals directly into an appropriate anchored interval (ie: only run the `element_parser` on the 1 endpoint)

PR #196, opened 2022-06-17
Tests, docs and some small fixes to IntervalSets

haberdashPI:
This adds some documentation for `IntervalSets` and a test for the breaking behavior that lead  us to revert #175.

Also @omus, was there a reason you removed `find_intervals` from the docs? That's definitely one of the more useful functions I'm leveraging from the original PR. (I saw earlier that you remvoed it from the list of exports; not sure I understand that bit either but at the time didn't see a problem with an initial PR being that way, but I had missed that it was also removed from docs, so I'm wondering if you thought it was an implementation detail).

Another few notes:

- I've made it possible for `IntervalSet`'s field to be any `AbstractVector`. There are some cases where I have a view of `TimeSpan.jl` arrays that I want to be treated as `Interval.jl` arrays, I use an AbstractArray that converts the objects on the fly during `getindex`.
- I've removed the call to `IntervalSet` in `find_intersections`. This function really is operating on arrays of intervals: it computes intersections between individual intervals and reports on which intervals in `y` intersect with intervals in `x`. No merging or re-ordering of the intervals should ever happen.

### DIFF
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -1,6 +1,12 @@
 
 ###### Set-related Helpers #####
 
+"""
+    IntervalSet(x::AbstractVector{<:AbstractInterval})
+
+Interpret an array of intervals as a set of points: the union of all points in the
+intervals.
+"""
 struct IntervalSet{T<:AbstractInterval}
     items::Vector{<:AbstractInterval}
 end
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -9,6 +9,8 @@ intervals. Set operations over them will return an IntervalSet containing the
 fewest number of intervals that can be used to represent the resulting point set. 
 Unbounded intervals are not supported.
 
+see also: https://en.wikipedia.org/wiki/Interval_arithmetic#Interval_operators
+
 ## Examples
 
 ```jldoctest
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -146,7 +146,7 @@ function unbunch((i, interval)::Tuple, tracking; lt=isless)
     return eltype[(i, LeftEndpoint(interval)), (i, RightEndpoint(interval))]
 end
 
-function unbunch(a::Union{AbstractVector{<:AbstractInterval}, AbstractIntervals, 
+function unbunch(a::Union{AbstractVector{<:AbstractInterval}, AbstractIntervals}, 
                  b::Union{AbstractVector{<:AbstractInterval}, AbstractIntervals}; kwargs...)
     tracking = endpoint_tracking(a, b)
     a_ = unbunch(a, tracking; kwargs...)
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -401,10 +401,10 @@ Base.intersect(x::IntervalSet, y::IntervalSet) = mergesets((inx, iny) -> inx && 
 Base.union(x::IntervalSet, y::IntervalSet) = mergesets((inx, iny) -> inx || iny, x, y)
 Base.setdiff(x::IntervalSet, y::IntervalSet) = mergesets((inx, iny) -> inx && !iny, x, y)
 Base.symdiff(x::IntervalSet, y::IntervalSet) = mergesets((inx, iny) -> inx ⊻ iny, x, y)
-Base.issubset(x::AbstractIntervalSets, y::AbstractIntervalSets) = isempty(setdiff(x, y))
-Base.isdisjoint(x::AbstractIntervalSets, y::AbstractIntervalSets) = isempty(intersect(x, y))
+Base.issubset(x::AbstractIntervals, y::AbstractIntervals) = isempty(setdiff(x, y))
+Base.isdisjoint(x::AbstractIntervals, y::AbstractIntervals) = isempty(intersect(x, y))
 
-function Base.issetequal(x::AbstractIntervalSets, y::AbstractIntervalSets)
+function Base.issetequal(x::AbstractIntervals, y::AbstractIntervals)
     x, y, tracking = unbunch(union(IntervalSet(x)), union(IntervalSet(y)))
     return x == y || all(isempty, bunch(x, tracking)) && all(isempty, bunch(y, tracking))
 end
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -63,17 +63,32 @@ Base.eltype(::IntervalSet{T}) where T = T
 Base.:(==)(a::IntervalSet, b::IntervalSet) = issetequal(a, b)
 Base.isequal(a::IntervalSet, b::IntervalSet) = isequal(a.items, b.items)
 Base.convert(::Type{T}, intervals::IntervalSet) where T <: AbstractArray = convert(T, intervals.items)
-function Base.show(io::IO, ::MIME"text/plain", x::IntervalSet) 
+function Base.show(io::Base.AbstractPipe, ::MIME"text/plain", x::IntervalSet) 
     intervals = union(x)
     iocompact = IOContext(io, :compact => true)
     print(io, "$(length(intervals))-interval ")
     show(io, MIME"text/plain"(), typeof(x))
     println(io, ":")
-    for interval in intervals.items[1:(end-1)]
-        show(iocompact, MIME"text/plain"(), interval)
-        println(io, "")
+    nrows = displaysize(io)[1]
+    half = fld(nrows, 2) - 2
+    if nrows ≥ length(intervals.items) && half > 1
+        for interval in intervals.items[1:(end-1)]
+            show(iocompact, MIME"text/plain"(), interval)
+            println(io, "")
+        end
+        isempty(intervals) || show(iocompact, MIME"text/plain"(), intervals.items[end])
+    else
+        for interval in intervals.items[1:half]
+            show(iocompact, MIME"text/plain"(), interval)
+            println(io, "")
+        end
+        println(io, "⋮")
+        for interval in intervals.items[(end-half+1):end-1]
+            show(iocompact, MIME"text/plain"(), interval)
+            println(io, "")
+        end
+        show(iocompact, MIME"text/plain"(), intervals.items[end])
     end
-    isempty(intervals) || show(iocompact, MIME"text/plain"(), intervals.items[end])
 end
 
 # currently (to avoid breaking changes) new methods for `Base`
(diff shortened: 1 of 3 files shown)


haberdashPI:
I'm confused about the error I'm getting for my doctests: seems like `..` should be in scope if I'm calling `using Intervals`.

src/interval_sets.jl
```diff
@@ -1,29 +1,76 @@
...
 
-Base.copy(intervals::IntervalSet{T}) where T = IntervalSet{T}(copy(intervals.items))
+Base.copy(intervals::IntervalSet{T, A}) where {T, A} = IntervalSet{T, A}(copy(intervals.items))
 Base.length(intervals::IntervalSet) = length(intervals.items)
 Base.iterate(intervals::IntervalSet, args...) = iterate(intervals.items, args...)
 Base.eltype(::IntervalSet{T}) where T = T
-Base.:(==)(a::IntervalSet, b::IntervalSet) = issetequal(a, b)
+Base.:(==)(a::IntervalSet, b::IntervalSet) = a.items == b.items
+Base.isequal(a::IntervalSet, b::IntervalSet) = isequal(a, b)
+Base.Array(intervals::IntervalSet) = intervals.items
```
rofinn, line 67: I don't believe we want to include this. This doesn't fit any existing `Array` constructor dispatch and in most cases you probably just want to call `collect`.

src/interval_sets.jl
```diff
@@ -1,29 +1,76 @@
...
 
-const AbstractIntervals = Union{AbstractInterval, IntervalSet}
+# currently (to avoid breaking changes) new methods for `Base`
+# accept `IntervalSet` objects and Interval singletons.
+const AbstractIntervalSets = Union{AbstractInterval, IntervalSet}
+# internal methods need to operate on both sets and arrays of intervals
+const AbstractIntervals = Union{AbstractIntervalSets, AbstractArray{<:AbstractInterval}}
```
rofinn, line 73: I'd like to move away from dispatching on vectors of intervals. I think we should make an explicit dispatch when we want to do that rather than having this union type floating around which risks introducing more breaking changes.

test/sets.jl
```diff
@@ -46,7 +50,7 @@ end
         @test isdisjoint(setdiff(a, b), b)
         @test !isdisjoint(a, a)
 
-        intersections = find_intersections(a, b)
+        intersections = find_intersections(Array(a), Array(b))
```
rofinn, line 53: Does `collect` not work here?

rofinn (changes requested):
This seems fine minus the addition of `AbstractVector{<:AbstractInterval}` to the `AbstractIntervals` union.

src/interval_sets.jl
haberdashPI: This goal is to avoid a copy. This seemed like the most natural name to access the underlying array. I've seen similar patterns used elsewhere. I guess it could also be called `parent`, but that seems like more of a pun.

test/sets.jl
haberdashPI: yes, though `Array` is faster since it avoids the copy.

src/interval_sets.jl
haberdashPI: One could just access `.items`, but I think having an function helps to communicate that this *could* avoid a copy but might involve some computation (e.g. if the internal representation of `IntervalSet` changes).

haberdashPI: I also think this is perfectly in line with the semantics of `Array` from base. The method constructs an array from its single argument.

test/sets.jl
omus: For this case you could just dispatch on `IntervalSet` to and avoid the copy that way

rofinn:
> I'm confused about the error I'm getting for my doctests: seems like .. should be in scope if I'm calling using Intervals.

I think you want to move the using statement to

```jldoctest; setup = :(using Intervals)
...
```

That being said, the output also doesn't match so it'll still fail. On a similar note, I think we should probably probably just implement `print` for this type, so you don't need to include the `convert` calls in the docstring examples. I can always fix this in my MR if you'd prefer?

src/interval_sets.jl
```diff
@@ -1,28 +1,73 @@
...
+
+see also: https://en.wikipedia.org/wiki/Interval_arithmetic#Interval_operators
+
+## Examples
+
+```jldoctest
+julia> using Intervals
+julia> convert(Array, union(IntervalSet(1..5), IntervalSet(3..8)))
+1-element Vector{Interval{Int64, Closed, Closed}}:
+ Interval{Int64, Closed, Closed}(1, 8)
```
omus, line 19: For these docstrings why not just have it display the `IntervalSet`?

src/interval_sets.jl
```diff
@@ -1,28 +1,73 @@
...
+ Interval{Int64, Closed, Open}(1, 4)
+ Interval{Int64, Open, Closed}(9, 10)
+```
 """
-struct IntervalSet{T<:AbstractInterval} <: AbstractSet{T}
-    # TODO: Use Dict internally once union!(Vector{AbstractInterval}) is fully deprecated
-    items::Vector{T}
+struct IntervalSet{I <: AbstractInterval, A <: AbstractVector{I}}
+    items::A
 end
```
omus, line 53: > I've made it possible for `IntervalSet`'s field to be any `AbstractVector`. There are some cases where I have a view of TimeSpan.jl arrays that I want to be treated as Interval.jl arrays, I use an `AbstractArray` that converts the objects on the fly during `getindex`.

I think this is most like a premature optimization. The internal representation of an `IntervalSet` could change in the future to no longer be ordered so we can't always guarantee that this will avoid a copy.

src/interval_sets.jl
```diff
@@ -1,28 +1,73 @@
...
-IntervalSet() = IntervalSet{AbstractInterval}(AbstractInterval[])
+IntervalSet() = IntervalSet(AbstractInterval[])
 
-Base.copy(intervals::IntervalSet{T}) where T = IntervalSet{T}(copy(intervals.items))
+Base.copy(intervals::IntervalSet{T, A}) where {T, A} = IntervalSet{T, A}(copy(intervals.items))
 Base.length(intervals::IntervalSet) = length(intervals.items)
 Base.iterate(intervals::IntervalSet, args...) = iterate(intervals.items, args...)
 Base.eltype(::IntervalSet{T}) where T = T
-Base.:(==)(a::IntervalSet, b::IntervalSet) = issetequal(a, b)
+Base.:(==)(a::IntervalSet, b::IntervalSet) = a.items == b.items
```
omus, line 65: Any reason for this change? I think `issetequal` is the right direction to go

src/interval_sets.jl
```diff
@@ -384,15 +425,15 @@ end
...
+        y::AbstractVector{<:AbstractInterval}
     )
 
 Returns a `Vector{Vector{Int}}` where the value at index `i` gives the indices to all
 intervals in `y` that intersect with `x[i]`.
 """
-function find_intersections(x_::AbstractIntervals, y_::AbstractIntervals)
-    xa, ya = IntervalSet(x_), IntervalSet(y_)
+function find_intersections(x_::Union{AbstractInterval, AbstractVector{<:AbstractInterval}}, y_::Union{AbstractInterval, AbstractVector{<:AbstractInterval}})
+    xa, ya = vcat(x_), vcat(y_)
```
omus, line 436: Use multiple dispatch to clean things up here:
```suggestion
find_intersections(x, y) = find_intersections(vcat(x), vcat(y))

function find_intersections(xa::AbstractVector{<:AbstractInterval}}, ya::AbstractVector{<:AbstractInterval}})
```

src/interval_sets.jl
omus: My vote is to define this as:
```suggestion
struct IntervalSet{T <: AbstractInterval}
    items::Vector{T}
end
```

src/interval_sets.jl
rofinn: Because currently IntervalSet doesn’t have nice printing. See comment above.

src/interval_sets.jl
haberdashPI: Fair enough. And actually the main reason I needed this no longer exists because it was mostly an issue when `IntervalSet` was called as a part of `find_intersections`, which I realized doesn't actually make sense there.

src/interval_sets.jl
haberdashPI: Agreed; I think this is stale code. I'll fix it.

src/interval_sets.jl
```diff
@@ -12,59 +12,54 @@ see also: https://en.wikipedia.org/wiki/Interval_arithmetic#Interval_operators
...
-Base.isequal(a::IntervalSet, b::IntervalSet) = isequal(a, b)
+Base.:(==)(a::IntervalSet, b::IntervalSet) = issetequal(a, b)
+Base.isequal(a::IntervalSet, b::IntervalSet) = isequal(a.items, b.items)
 Base.convert(::Type{T}, intervals::IntervalSet) where T <: AbstractArray = convert(T, intervals.items)
+function Base.show(io::IO, ::MIME"text/plain", x::IntervalSet) 
+    intervals = union(x)
+    iocompact = IOContext(io, :compact => true)
+    for interval in intervals.items[1:(end-1)]
+        show(iocompact, MIME"text/plain"(), interval)
+        print(io, " ∪ ")
```
rofinn, line 59: This seems a bit cutesy, but I suppose it's a nice visual identifier that you aren't working with an `AbstractVector` or `AbstractSet`?

rofinn (approved):
LGTM

src/interval_sets.jl
haberdashPI: Yeah, I think it clearly identifies it as an interval set, and is pretty readable. I also thought of spelling out `union([ [0 .. 1], [2 .. 4) ])`  but the additional `[]` brackets around everything seemed harder to read and without them it suggests `union` can be called with more than a pair of IntervalSet arguments, which doesn't seem quite right either.

rofinn: I was just thinking it would just look like:
```
IntervalSet{Interval{Int}} with 3 elements:
  (0 .. 2]
  [4 .. 6]
  [8 .. 11)
```
🤷🏻 Does @omus have an opinion?

rofinn: If there isn't any feedback by tomorrow I'll just merge this as is.

haberdashPI: That alternative seems reasonable, and probably more sane for larger sets. I'll switch to that.

haberdashPI:
I think the failures here are about how types get printed in 1.6 vs. 1.7, so I believe this is good to go??

PR #201, opened 2022-10-06
Handle empty-interval edge case in `find_intersections`

### LINKED ISSUE #200: stackoverflow from recursive call in `find_intersections`
It is very easy to cause a stackoverflow from:

https://github.com/invenia/Intervals.jl/blob/a785b56ce5e0091928fbb152bd2aa4834ada00d4/src/interval_sets.jl#L480

In my case I had TimeSpans because I forgot to convert them to intervals first.

I believe the intention of this code is to `collect` any iterators so we get vectors to hit the next method, https://github.com/invenia/Intervals.jl/blob/a785b56ce5e0091928fbb152bd2aa4834ada00d4/src/interval_sets.jl#L481. (I'm not sure why 1-arg `vcat` would be used over collect but I don't think that's the problem here). 

However:

* dispatching on the eltype of a container [isn't good practice](https://github.com/jrevels/YASGuide#programming-guidelines), because valid containers like `Any[Interval(1,2)]` won't get hit (and `vcat` doesn't help).
* if, like in my case, you don't have valid intervals, you should get an error, not a stackoverflow

This is extra annoying on MacOS because of the longstanding Julia issue where stackoverflows don't error properly on MacOS.

The easiest fix for the stackoverflow would be to remove the first method, however that would break any code that relies on `find_intersections` doing this collecting for them. I think a fix that preserves this behavior could look like:
```julia
find_intersections(x, y) = _find_intersections(collect(x), collect(y))
find_intersections(x::AbstractVector{<:AbstractInterval}, y::AbstractVector{<:AbstractInterval})  =  _find_intersections(x, y)
```
and then defining
```julia
function _find_intersections(x,y)
   ...
end
```
with what is currently in the method at https://github.com/invenia/Intervals.jl/blob/a785b56ce5e0091928fbb152bd2aa4834ada00d4/src/interval_sets.jl#L481.

haberdashPI:
Closes #200 

Title says it all. This fix also will address some edge cases where `find_intersections` would throw a StackOverflowError when passed an incorrect type.

Small aside: When I ran tests locally there was a failure on 1.8 (had to run on 1.7). Looks to be related to a now closed issue in 1.8.

### DIFF
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -471,11 +471,12 @@ end
 """
     find_intersections(x, y)
 
-Returns a `Vector{Vector{Int}}` where the value at index `i` gives the indices to all
-intervals in `y` that intersect with `x[i]`. Calls collect on the arguments if they
-aren't already arrays.
+Find the intervals in `y` that intersect with intervals in `x`. Returns a vector, `z`, where
+each element `z[i]` is a vector of all indices in `y` that intersect with `x[i]`. The values
+`x` and `y` should be iterables of intervals. 
 """
 find_intersections(x, y) = find_intersections_(collect(x), collect(y))
+find_intersections(x::AbstractVector, y::Abstractvector) = find_intersections_(x, y)
 function find_intersections_(x::AbstractVector, y::AbstractVector)
     (isempty(x) || isempty(y)) && return Vector{Int}[]
     tracking = endpoint_tracking(x, y)
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -476,7 +476,7 @@ intervals in `y` that intersect with `x[i]`. Calls collect on the arguments if t
 aren't already arrays.
 """
 find_intersections(x, y) = find_intersections_(collect(x), collect(y))
-function find_intersections_(x::AbstractVector{<:AbstractInterval}, y::AbstractVector{<:AbstractInterval})
+function find_intersections_(x::AbstractVector, y::AbstractVector)
     (isempty(x) || isempty(y)) && return Vector{Int}[]
     tracking = endpoint_tracking(x, y)
     lt = intersection_isless_fn(tracking)
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -469,22 +469,13 @@ function intersection_isless_fn(::TrackEachEndpoint)
 end
 
 """
-    find_intersections(
-        x::AbstractVector{<:AbstractInterval},
-        y::AbstractVector{<:AbstractInterval}
-    )
+    find_intersections(x, y)
 
 Returns a `Vector{Vector{Int}}` where the value at index `i` gives the indices to all
-intervals in `y` that intersect with `x[i]`.
+intervals in `y` that intersect with `x[i]`. Calls collect on the arguments if they
+aren't already arrays.
 """
-find_intersections(x, y) = find_intersections_(vcat(x), vcat(y))
-function find_intersections_(x, y)
-    if isempty(x) && isempty(y)
-        return Vector{Int}[]
-    else
-        throw(ArgumentError("Expected arrays of `AbstractInterval` objects"))
-    end
-end
+find_intersections(x, y) = find_intersections_(collect(x), collect(y))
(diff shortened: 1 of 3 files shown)


src/interval_sets.jl
```diff
@@ -477,8 +477,16 @@ end
 Returns a `Vector{Vector{Int}}` where the value at index `i` gives the indices to all
 intervals in `y` that intersect with `x[i]`.
 """
-find_intersections(x, y) = find_intersections(vcat(x), vcat(y))
-function find_intersections(x::AbstractVector{<:AbstractInterval}, y::AbstractVector{<:AbstractInterval})
+find_intersections(x, y) = find_intersections_(vcat(x), vcat(y))
```
ericphanson, line 480: Why 1-arg vcat?
```suggestion
find_intersections(x, y) = find_intersections_(collect(x), collect(y))
```

ericphanson (commented):
closes https://github.com/invenia/Intervals.jl/issues/200?

I suggested arranging the dispatches slightly differently there, which I think has some advantages. With this version:

* Empty checks are repeated in two methods
* errors depend on contents, not type: if the array has the wrong type and is empty, no error. Non-empty, and error

src/interval_sets.jl
haberdashPI: To support single intervals in one or both arguments, e.g. `find_intersections([1..2, 4..5], 1.5..4.5)`

haberdashPI: I'll add a comment to clarify

haberdashPI: As I'm thinking through this more, I think this is far less useful than I thought it was before 😆, `collect` seems better to me now.

src/interval_sets.jl
```diff
@@ -469,16 +469,15 @@ function intersection_isless_fn(::TrackEachEndpoint)
...
 Returns a `Vector{Vector{Int}}` where the value at index `i` gives the indices to all
-intervals in `y` that intersect with `x[i]`.
+intervals in `y` that intersect with `x[i]`. Calls collect on the arguments if they
+aren't already arrays.
 """
-find_intersections(x, y) = find_intersections(vcat(x), vcat(y))
-function find_intersections(x::AbstractVector{<:AbstractInterval}, y::AbstractVector{<:AbstractInterval})
+find_intersections(x, y) = find_intersections_(collect(x), collect(y))
+function find_intersections_(x::AbstractVector, y::AbstractVector)
+    (isempty(x) || isempty(y)) && return Vector{Int}[]
```
omus, line 480: May as well avoid collecting types that are already `AbstractVector`s such as ranges:
```suggestion
find_intersections(x, y) = find_intersections(collect(x), collect(y))
function find_intersections(x::AbstractVector, y::AbstractVector)
    (isempty(x) || isempty(y)) && return Vector{Int}[]
```

src/interval_sets.jl
```diff
@@ -469,16 +469,15 @@ function intersection_isless_fn(::TrackEachEndpoint)
 end
 
 """
-    find_intersections(
-        x::AbstractVector{<:AbstractInterval},
-        y::AbstractVector{<:AbstractInterval}
-    )
+    find_intersections(x, y)
```
omus, line 472: The docstring should go into more detail about acceptable arguments to this function now that context from the types is no longer available

test/anchoredinterval.jl
```diff
@@ -583,7 +583,7 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
             r = r[1:5]
 
             # https://github.com/JuliaLang/julia/issues/33882
-            @test r isa StepRangeLen
+            @test r isa (VERSION < v"1.8" ? StepRangeLen : StepRange)
```
omus, line 586: Best to make these fixes as separate PRs so they can be merged faster

omus: After looking into it I think this fix is incorrect. I've made a new fix in #204
```suggestion
            @test r isa StepRangeLen
```

src/interval_sets.jl
haberdashPI: If `collect` is poorly implemented the proposed modification could lead to a stack overflow. I've added a new method, and kept `find_intersections_`

Issue #199, opened 2022-09-09
StepRangeLen broken for AnchoredIntervals on Julia 1.8


BaileyBox:
`StepRange` has added the ability to also use `Length` as a parameter in Julia1.8 (along with making `stop` and `length` as only constructors possible). This change somehow broke `StepRangeLen` with `AnchoredIntervals`. 

An example showing how this is broken:

Julia1.6:
```julia
using Intervals

s = AnchoredInterval{-1}(1):1:AnchoredInterval{-1}(5)
r = s[1:5]

s isa StepRangeLen # false
s isa StepRange # true
r isa StepRangeLen # true
r isa StepRange # false
```

Julia1.8:
```julia
using Intervals

s = AnchoredInterval{-1}(1):1:AnchoredInterval{-1}(5)
r = s[1:5]

s isa StepRangeLen # false
s isa StepRange # true
r isa StepRangeLen # false
r isa StepRange # true
```

This failure can also be observed in the `CI` and `Nightly` pipelines. Current test failing is at `anchoredinterval.jl:586`.

PR #203, opened 2022-10-13
Optimize `find_intersections` for closed intervals

ericphanson:
This adds a method for `find_intersections` with closed-endpoints which is *much* faster, at least in my example problem for which the current performance too slow. The algorithm is just based on sorting and binary searching, and is based on some code that I wrote for a beacon-internal project last week which I didn’t realize was actually in fact basically a solution for this problem.

This speeds up the problem in https://github.com/beacon-biosignals/DataFrameIntervals.jl/issues/24 from 22s to 1s (with `N_POINTS = 10^5` and `N_INTERVALS = 1000`), and more importantly to me, the `N_POINTS = 10^6` and `N_INTERVALS = 3000` version now takes 35s. I don't have a timing with the current code on that exact problem, but a very similar one (off which the MWE was based) took 2.7 hours.

This is covered by the current tests (the ones with closed-closed intersections), but I can add specific tests if someone thinks something in particular would be good to cover.

This algorithm works very differently depending on which argument comes first, but I get similar performance (on my example problem at least) no matter which argument is first. The number of individual allocations does change by a lot, however:
```julia
julia> @time find_intersections(L.interval, R.interval);
  4.324992 seconds (98.94 k allocations: 1.765 GiB, 1.87% gc time)

julia> @time find_intersections(R.interval, L.interval);
  4.504758 seconds (11.01 M allocations: 1.834 GiB, 1.95% gc time)
```
(this is on the `N_POINTS = 10^6` and `N_INTERVALS = 3000` version of the example problem; the timings ~5s instead of 35s because this is only the intersection portion; now the DataFramesIntervals overhead dominates, whereas it used to be the `find_intersections` time).

I am not sure this is so much faster in all cases, or only this particular problem. It however also speeds up the case of `interval_join(L, L; on=:interval, makeunique=true)` from 22s to 8s. This case is a very different shape than the `L, R` case, so since it speeds up both cases I suspect it's just a faster appraoch. However if someone suggests other problems I can try them too.

---

I only did the closed-closed case because I didn't want to think about all the endpoints. I would appreciate if we could leave other endpoints to future work, since this should already be a big improvement in some cases at least.

### DIFF
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -522,21 +522,25 @@ function find_intersections_helper!(result, x, y, lt)
     return unique!.(result)
 end
 
+
 function find_intersections(x::AbstractVector{<:AbstractInterval{T1, Closed, Closed}}, y::AbstractVector{<:AbstractInterval{T2, Closed, Closed}}) where {T1, T2}
     # Strategy:
     # two binary searches per interval `I` in `x`
     # * identify the set of intervals in `y` that start during-or-after `I`
     # * identify the set of intervals in `y` that stop before-or-during `I`
     # * intersect them
-    @time begin
-        starts = first.(y)
-        starts_perm = sortperm(starts)
-        starts_sorted = starts[starts_perm]
+    starts = first.(y)
+    starts_perm = sortperm(starts)
+    starts_sorted = starts[starts_perm]
 
-        stops = last.(y)
-        stops_perm = sortperm(stops)
-        stops_sorted = stops[stops_perm]
-    end
+    # Sneaky performance optimization (makes a huge difference!)
+    # Rather than sorting `stops` relative to `y`, we sort it relative to `starts`.
+    # This allows us to work in the "starts" from of reference until the very end.
+    # In particular, when we intersect the sets of intervals obtained from starts and from stops,
+    # the "starts" set can be kept as a `UnitRange`, making the intersection *much* faster.
+    stops = last.(y[starts_perm])
+    stops_perm = sortperm(stops)
+    stops_sorted = stops[stops_perm]
 
     results = Vector{Vector{Int}}(undef, length(x))
     for (i, I) in enumerate(x)
@@ -545,22 +549,25 @@ function find_intersections(x::AbstractVector{<:AbstractInterval{T1, Closed, Clo
         if idx_first < 1
             results[i] = Int[]
             continue
-            # return Int[]
         end
-        starts_before_or_during = @view starts_perm[1:idx_first]
 
         # find all the stops which occur after the start of `I`
         idx_last = searchsortedfirst(stops_sorted, first(I))
 
         if idx_last > length(stops_sorted)
             results[i] = Int[]
-        continue
-            # return Int[]
+            continue
         end
-        stops_during_or_after = @view stops_perm[idx_last:end]
+
+        # Working in "starts" frame of reference
+        starts_before_or_during = 1:idx_first
+        stops_during_or_after = stops_perm[idx_last:end]
 
         # Intersect them
-        results[i] = intersect(starts_before_or_during, stops_during_or_after)
+        r = intersect(starts_before_or_during, stops_during_or_after)
+
+        # *Now* go back to y's sorting order, post-intersection.
+        results[i] = starts_perm[r]
     end
     return results
 end
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -488,6 +488,40 @@ function find_intersections(x::AbstractVector{<:AbstractInterval}, y::AbstractVe
     return find_intersections_helper!(result, x_endpoints, y_endpoints, lt)
 end
 
+function find_intersections(x::AbstractVector{<:AbstractInterval{T, Closed, Closed}}, y::AbstractVector{<:AbstractInterval{T, Closed, Closed}}) where {T}
+    yperm_first = sortperm(y; by=first)
+    yperm_last = sortperm(y; by=last)
+    y_sorted_first = y[yperm_first]
+    y_sorted_last = y[yperm_last]
+
+    starts = first.(y_sorted_first)
+    stops = last.(y_sorted_last)
+
+    results = map(x) do I
+
+        # starts before or during the interval. That means find all points whos START
+        # is less than the END of the interval
+        idx_first = searchsortedlast(starts, last(I))
+        if idx_first < 1
+            starts_before_or_during = 1:0
+        else
+            starts_before_or_during = yperm_first[1:idx_first]
+        end
+
+        idx_last = searchsortedfirst(stops, first(I))
+
+        if idx_last > length(stops)
+            stops_during_or_after = 1:0
+        else
+            stops_during_or_after = yperm_last[idx_last:end]
+        end
+
+        return intersect(starts_before_or_during, stops_during_or_after)
+    end
+    return results
+end
+
+
 function find_intersections_helper!(result, x, y, lt)
     active_xs = Set{Int}()
     active_ys = Set{Int}()
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -522,7 +522,10 @@ function find_intersections_helper!(result, x, y, lt)
     return unique!.(result)
 end
 
-function find_intersections(x::AbstractVector{<:AbstractInterval{T1, Closed, Closed}}, y::AbstractVector{<:AbstractInterval{T2, Closed, Closed}}) where {T1, T2}
+function find_intersections(
+    x::AbstractVector{<:AbstractInterval{T1,Closed,Closed}},
+    y::AbstractVector{<:AbstractInterval{T2,Closed,Closed}},
+) where {T1,T2}
     # Strategy:
     # two binary searches per interval `I` in `x`
     # * identify the set of intervals in `y` that start during-or-after `I`
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -547,14 +547,14 @@ function find_intersections(
 
     results = Vector{Vector{Int}}(undef, length(x))
     for (i, I) in enumerate(x)
-        # find all the starts which occur before the end of `I`
+        # find all the starts which occur before or at the end of `I`
         idx_first = searchsortedlast(starts_sorted, last(I); lt=(<))
         if idx_first < 1
             results[i] = Int[]
             continue
         end
 
-        # find all the stops which occur after the start of `I`
+        # find all the stops which occur at or after the start of `I`
         idx_last = searchsortedfirst(stops_sorted, first(I); lt=(<))
         if idx_last > len
             results[i] = Int[]
@@ -563,7 +563,7 @@ function find_intersections(
 
         # Working in "starts" frame of reference
         starts_before_or_during = 1:idx_first
-        stops_during_or_after = @views stops_perm[idx_last:end]
+        stops_during_or_after = @view stops_perm[idx_last:end]
 
         # Intersect them
         r = intersect(starts_before_or_during, stops_during_or_after)
--- a/src/interval_sets.jl
+++ b/src/interval_sets.jl
@@ -548,15 +548,14 @@ function find_intersections(
     results = Vector{Vector{Int}}(undef, length(x))
     for (i, I) in enumerate(x)
         # find all the starts which occur before the end of `I`
-        idx_first = searchsortedlast(starts_sorted, last(I))
+        idx_first = searchsortedlast(starts_sorted, last(I); lt=(<))
         if idx_first < 1
             results[i] = Int[]
             continue
         end
 
         # find all the stops which occur after the start of `I`
-        idx_last = searchsortedfirst(stops_sorted, first(I))
-
+        idx_last = searchsortedfirst(stops_sorted, first(I); lt=(<))
         if idx_last > len
             results[i] = Int[]
             continue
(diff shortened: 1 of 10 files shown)


src/interval_sets.jl
```diff
@@ -521,3 +521,55 @@ function find_intersections_helper!(result, x, y, lt)
...
+    # * identify the set of intervals in `y` that start during-or-after `I`
+    # * identify the set of intervals in `y` that stop before-or-during `I`
+    # * intersect them
+    starts = first.(y)
+    starts_perm = sortperm(starts)
+    starts_sorted = starts[starts_perm]
+
+    # Sneaky performance optimization (makes a huge difference!)
+    # Rather than sorting `stops` relative to `y`, we sort it relative to `starts`.
+    # This allows us to work in the "starts" from of reference until the very end.
```
ericphanson, line 540:
```suggestion
    # This allows us to work in the "starts" frame of reference until the very end.
```

ericphanson:
CI failures are https://github.com/invenia/Intervals.jl/issues/199

haberdashPI:
Awesome! The commenting really helped. I think this is pretty clear, and I didn't notice any bugs.

So, want to make sure I understand this correctly. Since I'd like to generalize this to all interval types at some point.

In essence what's going on is that rather than scanning along for each interval that intersects with A, you can sorted-search for the last such interval, get its index, and then compute intersections in the UnitRange domain. Cool!

In principle you could have just 1 such interval in A per interval in B so I think the complexity class remains: O(N log N) + O(M log M) + O(K log max(N,M)) where N and M are the lengths of A and B interval sets and M the length of the output. However for use cases where there are many intervals in B for the intervals in A (or vice versa) this will be faster and that's actually a pretty common use case!

ericphanson:
> In essence what's going on is that rather than scanning along for each interval that intersects with A, you can sorted-search for the last such interval, get its index, and then compute intersections in the UnitRange domain. Cool!

Yep!

> O(N log N) + O(M log M) + O(K log max(N,M)) where N and M are the lengths of A and B interval sets and M the length of the output

What is K here? The length of the flattened output?

I think this sounds roughly right. I think of it as: we need to sort `B`, so that's `M*log(M)`, then for each element of `A` we do `log(M)` work in binary searching, so that's `N*log(M)`, plus some amount of work for the number of intersections we actually find (to do our [permuting](https://github.com/invenia/Intervals.jl/pull/203/files#diff-104262f55cc91ffa6259fa23de615d6b7703559750681261584bb5ef891db421R572)). I guess that adds up to `K`, once we sum over all elements in A, if `K` is the total number of intersections (length of flattened output). So overall, `M*log(M) + N*log(M) + K`, I think, taking `K` as the total number of intersections. Does that seem right?

haberdashPI:
Yes, by `K` I mean the length of the output.

Cool, that makes sense. I think that should be enough for me to work on a version that generalizes to all interval types.

https://github.com/JuliaCollections/SortingAlgorithms.jl would be another option if we want to use a radix sort.

ericphanson:
I think we can just wait for the one in Base; sorting was not the bottleneck when I was profiling this.

If you were able to give it a look, could you do a review here? That might help get it in

src/interval_sets.jl
```diff
@@ -521,3 +521,55 @@ function find_intersections_helper!(result, x, y, lt)
...
+    # * intersect them
+    starts = first.(y)
+    starts_perm = sortperm(starts)
+    starts_sorted = starts[starts_perm]
+
+    # Sneaky performance optimization (makes a huge difference!)
+    # Rather than sorting `stops` relative to `y`, we sort it relative to `starts`.
+    # This allows us to work in the "starts" frame of reference until the very end.
+    # In particular, when we intersect the sets of intervals obtained from starts and from stops,
+    # the "starts" set can be kept as a `UnitRange`, making the intersection *much* faster.
```
omus, line 542: Any reason for using double-quote for `starts`?
```suggestion
    # Sneaky performance optimization (makes a huge difference!)
    # Rather than sorting `stops` relative to `y`, we sort it relative to `starts`.
    # This allows us to work in the `starts` frame of reference until the very end.
    # In particular, when we intersect the sets of intervals obtained from starts and from stops,
    # the `starts` set can be kept as a `UnitRange`, making the intersection *much* faster.
```

src/interval_sets.jl
```diff
@@ -521,3 +521,55 @@ function find_intersections_helper!(result, x, y, lt)
...
+        idx_first = searchsortedlast(starts_sorted, last(I))
+        if idx_first < 1
+            results[i] = Int[]
+            continue
+        end
+
+        # find all the stops which occur after the start of `I`
+        idx_last = searchsortedfirst(stops_sorted, first(I))
+
+        if idx_last > length(stops_sorted)
```
omus, line 559: Consider setting `length(stops_sorted) to a variable outside the loop

src/interval_sets.jl
```diff
@@ -521,3 +521,55 @@ function find_intersections_helper!(result, x, y, lt)
 
     return unique!.(result)
 end
+
+function find_intersections(
+    x::AbstractVector{<:AbstractInterval{T1,Closed,Closed}},
+    y::AbstractVector{<:AbstractInterval{T2,Closed,Closed}},
+) where {T1,T2}
```
omus, line 528: Currently only executed by a single test. That makes me worried... I'll see what I can do about this

src/interval_sets.jl
ericphanson: Done

ericphanson:
On the Julia slack (`#performance-helpdesk`), @miguelraz pointed out:

>  couldn't you sort the x by the last(I), and if any of the idx_last > len then you know the search is done

I think that could speed things up a lot in the case where all of the `y`'s are to the left of many of the `x`s, since once we find the first such case, we know the rest are empty.

I think we could also go the other way, sorting `x` by `first`, to handle the case in which all of the `y`'s are to the right of many of the `x`'s.

I think a generalization of this is: if the `x`'s are sorted by `first`, then as we iterate through the loop over x, we know `idx_last ` should be monotone-increasing, and we can do shorter and shorter binary searches by starting after the previous `idx_last`. Likewise, if the `x`s were sorted by `last`, `idx_first` should be monotone-decreasing. Then we recover the "all to the right" and "all to the left" cases by noticing once `idx_first` or `idx_last` becomes extremal.

I am not quite sure if we could somehow use two sorted copies of `x` like we do with `y`, in order to get the benefits of both sortings. I think perhaps we could by delaying the intersections. In other words, we first compute: for each `x_start`, what are all the `y_stop`s that occur after it, and for each `x_stop`, what are all the `y_start`s that occur before it. We should be able to work in sorted coordinates such that these are all `UnitInterval`s, so they aren't too heavy.  Each of these lists is in a different order, coresponding to the sort-x-by-first and sort-x-by-last orders. Then we un-permute both orders to align back to the original order of `x`, and perform the intersections there.

This will be slightly more allocation heavy, but not incredibly so: it means we need to additionally allocate 2 `UnitRange`s per element of `x`, but that's a lot better than say 2 `Vector`s. We will also have a few more copies of `x` lying around, due to the additional sorting and permutation vectors. We can allocate everything upfront however, so we don't need to allocate in the loops themselves. I think this is probably worth it. It should be much faster in cases where many `x`s don't intersect with many `y`s, which I am guessing is most of the time that folks use `find_intersections`. At least in my use cases, I often have small intervals (one second or less) spread over hours.

ericphanson:
I tried the above approach in https://github.com/ericphanson/Intervals.jl/compare/eph/faster-intersections...ericphanson:Intervals.jl:eph/optimize-find-intersections-2. Disappointingly, it doesn't seem much faster, even when I try to shift around the intervals' ranges to better hit these optimizations:

```julia
using DataFrames, StableRNGs, Intervals
using Intervals: find_intersections

function make_example(n_intervals, n_points; point_max=10_000, interval_max=10_000)
    rng = StableRNG(123)
    points = [rand(rng, 1:point_max) for _ in 1:n_points]
    L = DataFrame(:point => points)
    transform!(L, :point => ByRow(p -> Interval{Int,Closed,Closed}(p, p)) => :interval)

    indices = map(1:n_intervals) do _
        i = rand(rng, 1:interval_max)
        return i:i+100
    end
    R = DataFrame(:indices => indices)
    transform!(R, :indices => ByRow(I -> Interval{Int,Closed,Closed}(first(I), last(I))) => :interval)
    return L, R
end

L, R = make_example(3000, 10^6; point_max=100_000, interval_max=10_000)


@time find_intersections(L.interval, R.interval);
@time find_intersections(R.interval, L.interval);
```

For this PR, I get 

```julia
julia> @time find_intersections(L.interval, R.interval);
  0.640732 seconds (2.01 M allocations: 256.068 MiB, 21.22% gc time)

julia> @time find_intersections(R.interval, L.interval);
  4.522413 seconds (72.02 k allocations: 294.140 MiB, 3.77% gc time)
```

For that branch, I get
```julia
julia> @time find_intersections(L.interval, R.interval);
  0.693489 seconds (2.01 M allocations: 317.866 MiB, 8.34% gc time)

julia> @time find_intersections(R.interval, L.interval);
  4.437335 seconds (72.04 k allocations: 294.323 MiB, 1.13% gc time)
```

src/interval_sets.jl
ericphanson: there's a few in `test/sets.jl`, def not just 1 (since they caught various bugs when developing this)

haberdashPI:
Wow... it's crazy to me that the first, most obvious way I thought of to make this a little more efficient was published in 2016! I would have expected it to be discovered before that... 

Looks like that paper might have some more optimizations to consider when we're ready to make this even faster.

src/interval_sets.jl
ericphanson: This method is called 6 times in the `test/sets.jl` file (I inserted a global, incremented it in this method, then checked before-vs-after running `test/sets.jl`.

ericphanson:
bump

haberdashPI:
I've made a few small changes here: https://github.com/ericphanson/Intervals.jl/pull/1

I believe this would be a simple way to get the faster algorithm working for all interval endpoints. (Since #205 is setup without including this change I have yet to running it on that more comprehensive set of tests). Just wanted to post here for transparency, I suspect that it should ultimatley become a PR to this version of the repo, once this and #205 merge.

ericphanson:
Yeah, I would really prefer we got this PR in as soon as possible and then incrementally expanded the functionality. We are currently forced to pirate this function in a lot of downstream code because the current performance is untenable.

ericphanson:
bump

omus:
Actively reviewing this PR today

src/interval_sets.jl
```diff
@@ -521,3 +521,56 @@ function find_intersections_helper!(result, x, y, lt)
...
+        idx_last = searchsortedfirst(stops_sorted, first(I))
+
+        if idx_last > len
+            results[i] = Int[]
+            continue
+        end
+
+        # Working in "starts" frame of reference
+        starts_before_or_during = 1:idx_first
+        stops_during_or_after = @views stops_perm[idx_last:end]
```
omus, line 567:
```suggestion
        stops_during_or_after = @view stops_perm[idx_last:end]
```

src/interval_sets.jl
```diff
@@ -521,3 +521,56 @@ function find_intersections_helper!(result, x, y, lt)
...
+        idx_first = searchsortedlast(starts_sorted, last(I))
+        if idx_first < 1
+            results[i] = Int[]
+            continue
+        end
+
+        # find all the stops which occur after the start of `I`
+        idx_last = searchsortedfirst(stops_sorted, first(I))
+
+        if idx_last > len
```
omus, line 560: Same as above:
```suggestion
        if idx_last == len + 1
```

src/interval_sets.jl
```diff
@@ -521,3 +521,56 @@ function find_intersections_helper!(result, x, y, lt)
...
+    # the `starts` set can be kept as a `UnitRange`, making the intersection *much* faster.
+    stops = last.(y[starts_perm])
+    stops_perm = sortperm(stops)
+    stops_sorted = stops[stops_perm]
+    len = length(stops_sorted)
+
+    results = Vector{Vector{Int}}(undef, length(x))
+    for (i, I) in enumerate(x)
+        # find all the starts which occur before the end of `I`
+        idx_first = searchsortedlast(starts_sorted, last(I))
```
omus, line 551: Required to pass new comparison tests in #205. Specifically when comparing `-0.0` and `0.0`
```suggestion
        idx_first = searchsortedlast(starts_sorted, last(I); lt=(<))
```

src/interval_sets.jl
```diff
@@ -521,3 +521,56 @@ function find_intersections_helper!(result, x, y, lt)
...
+        if idx_first < 1
```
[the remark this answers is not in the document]
omus: I'm just going to resolve these. I like the code as it is

omus (approved):

omus:
@haberdashPI pointed out a private PR which generalizes this for all intervals. I'll make a follow up PR for that which will be the 1.10.0 release

omus:
Re-ran the [benchmark posted above](https://github.com/invenia/Intervals.jl/pull/203#issuecomment-1283895623).

With the `lt=<` fix (7c31a48):
```julia
julia> @time find_intersections(L.interval, R.interval);
  0.655631 seconds (2.64 M allocations: 287.892 MiB, 6.09% gc time, 20.41% compilation time)

julia> @time find_intersections(R.interval, L.interval);
  5.724341 seconds (72.02 k allocations: 294.140 MiB, 0.94% gc time)
```

This PR before the fix (7294a569)
```julia
julia> @time find_intersections(L.interval, R.interval);
  0.686747 seconds (2.61 M allocations: 286.352 MiB, 7.78% gc time, 19.47% compilation time)

julia> @time find_intersections(R.interval, L.interval);
  5.218349 seconds (72.02 k allocations: 294.140 MiB, 0.85% gc time)
```

PR #211, opened 2023-01-25
Loosen type signatures to support UTCDateTimes

rofinn:


oxinabox (approved):

src/anchoredinterval.jl
```diff
@@ -386,8 +386,8 @@ canonicalize(target_type::Type{P}, p::P) where P <: Period = p
 
 ##### TIME ZONES #####
 
-function TimeZones.astimezone(i::AnchoredInterval{P, ZonedDateTime, L, R}, tz::TimeZone) where {P,L,R}
+function TimeZones.astimezone(i::AnchoredInterval{P, T, L, R}, tz::TimeZone) where {P,T,L,R}
     return AnchoredInterval{P, ZonedDateTime, L, R}(astimezone(anchor(i), tz))
```
iamed2, line 390:
```suggestion
    zdt = astimezone(anchor(i), tz)
    return AnchoredInterval{P, typeof(zdt), L, R}(zdt)
```
This could be done in the future. Probably this type of change would have to happen in a bunch of places for it to work and there's no use case right now, though, so we can hold off.

iamed2 (approved):

src/anchoredinterval.jl
rofinn: Added an issue for it. Might want to do that at the same time that we try to make TimeZones an optional dependency.

https://github.com/invenia/Intervals.jl/issues/212

omus (commented):
Post-review: seems reasonable. As an aside these TZ specific methods have always been a little weird

PR #214, opened 2023-04-07
Drop all v1 deprecations for v2

rofinn:
Removes all deprecations placed in src/deprecations and fixes/deletes any relevant tests. Commits are split up by types of deprecations being dropped.

@omus I can't seem to set you as a reviewer, but I'm guessing you're the primary person who should review these changes?


omus:
I've allocated some bandwidth to review this tomorrow

test/comparisons.jl
```diff
@@ -280,7 +274,7 @@ end
                 @test setdiff(IntervalSet(later), IntervalSet(earlier)) == IntervalSet(expected_xor[2:2])
 
                 # TODO: Sometimes expected_xor would get mutated in this call
-                @test symdiff([earlier], [later]) == expected_xor != union(expected_xor)
+                @test symdiff([earlier], [later]) == expected_xor == union(expected_xor)
```
omus, line 277: This minor change is generating some major internal debate. In looking over this series of tests again the intent is to show that the `IntervalSet` behaviour works as expected and that the vector of `Interval` may or may not match. It appears that a commit you added (e122bfe) introduced the additional check to see what the resulting value actually is which makes the intent less clear.

In looking over that commit is seems you just cared about having a check to see what the value actually is so by following what I original intended and your intent we should just do:
```suggestion
                @test symdiff([earlier], [later]) == union(expected_xor)
```

test/comparisons.jl
```diff
@@ -368,7 +360,7 @@ end
                 @test setdiff([later], [earlier]) == expected_xor[2:2]
                 @test setdiff(IntervalSet(later), IntervalSet(earlier)) == IntervalSet(expected_xor[2:2])
 
-                @test symdiff([earlier], [later]) == expected_xor != union(expected_xor)
+                @test symdiff([earlier], [later]) == expected_xor == union(expected_xor)
```
omus, line 363:
```suggestion
                @test symdiff([earlier], [later]) == union(expected_xor)
```

test/comparisons.jl
```diff
@@ -87,8 +87,6 @@ end
             @test_throws MethodError setdiff(later, earlier)
             @test_throws MethodError symdiff(earlier, later)
 
-            # Using a vector of intervals as sets
-            @test union([earlier, later]) == [earlier, later]
```
omus, line 91: These tests should stay as they serve to showcase why you need `IntervalSet` when dealing with sets of `Interval`s. The comment should be updated to say:
```
# Test using a vector of intervals as a set only for comparison purposes
```
Unfortunately, I'm unable to suggest these changes.

test/interval.jl
```diff
@@ -751,57 +741,42 @@
...
-        buffer = IOBuffer(
-            SERIALIZED_HEADER *
-            "\x004\x10\x01\bInterval\x1f\v՞\x84\xec\xf7-`\x87\xbbS\xe1Á\x88A\xd8\x01\t" *
-            "IntervalsD\x01\0\0\0\0\b\xe0\xe14\x10\x01\vInclusivity\x1f\v՞\x84\xec\xf7" *
-            "-`\x87\xbbS\xe1Á\x88A\xd8,\x02\0DML"
-        )
-
-        interval = deserialize(buffer)
-        @test interval isa Interval
-        @test interval == Interval{Closed,Open}(1, 2)
```
omus, line 804: Same as other deserialization comment

omus (changes requested):

test/anchoredinterval.jl
```diff
@@ -774,21 +766,6 @@ using Intervals: Bounded, Ending, Beginning, canonicalize, isunbounded
...
-            "\x004\x10\x01\x10AnchoredInterval\x1f\v՞\x84\xec\xf7-`\x87\xbb" *
-            "S\xe1Á\x88A\xd8\x01\tIntervalsD\x02\0\0\x001\xff\xff\xff\xff\0\b\xe14\x10" *
-            "\x01\vInclusivity\x1f\v՞\x84\xec\xf7-`\x87\xbbS\xe1Á\x88A\xd8,\x02\0DML"
-        )
-
-        interval = deserialize(buffer)
-        @test interval isa AnchoredInterval
-        @test interval == AnchoredInterval{-1,Int,Closed,Open}(2)
-    end
-
```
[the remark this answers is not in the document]
rofinn: Since this is already going into the break v2-DEV release I'm inclined to just drop the compat.jl code. Also, deserialization will likely break again with #217 anyway. If we don't drop it for v2 are we just going to retain this compat code indefinitely?

test/comparisons.jl
rofinn: So you want to keep the test just to test the `Base` behaviour... and test the negation?

test/comparisons.jl
rofinn: > resulting value actually is which makes the intent less clear.

Why is it less clear? The issue with the original check is that we were testing with multiple overloaded methods. If you were to read these tests without context you might be confused about why they don't match because our implementation of `union` for vectors of intervals is a bit weird. It seemed prudent to add a more explicit check for when we deprecated the `union` method. 

I suppose they could have been separate checks, but I like that the change is all on one line. Anyway, I agree that going forward we don't need to check both anymore, since `union` works as expected now.

rofinn (commented):
Okay, I've reviewed and while I agree with most of the comments I'd like to confirm that we want to:

1. Do we want to maintain v1 deserialization in v2 (it'll be a lot of work/bloat for all the proposed breaking changes)?
2. Do we want to include tests of Base behaviour (ie: that `union(intervalsets) != union(vectors)`)?

test/anchoredinterval.jl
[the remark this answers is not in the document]
omus: I agree we can't keep the serialization compat code forever but it is nice to have some compatibility with old serialized versions. I think having backwards compatible serialization with the latest previous version would be a good goal to have (1.9.0) in this case.

I'd personally probably include this compat change in this PR and then just remove it in #217 just to keep in the commit history.

test/comparisons.jl
omus: When we added the `IntervalSet` set methods tests we also added tests against vectors of intervals just to document any deviations in the behaviour but ultimately we didn't care what the result was. 

This originally looked like:
```julia
@test symdiff([earlier], [later]) != union(expected_xor)
@test symdiff(IntervalSet(earlier), IntervalSet(later)) == union(IntervalSet(expected_xor))
```
The vector of intervals test is just mirroring the `IntervalSet` test below but adjusts the comparison to the current behaviour

> Anyway, I agree that going forward we don't need to check both anymore, since union works as expected now.

Sounds good.

test/comparisons.jl
omus: Correct. For the same reason we test [vectors of intervals elsewhere](https://github.com/invenia/Intervals.jl/pull/214#discussion_r1206221922). Unless we can stop end users from using vectors of intervals seeing weird behaviour it's good to document/record the behaviour of vectors of intervals with the `Base` set functions.

omus:
> 1. Do we want to maintain v1 deserialization in v2 (it'll be a lot of work/bloat for all the proposed breaking changes)?

I'd make an issue against the v2 milestone and deal with this before we make the official v2.0.0 release. I don't think we need to keep deserialization compatibility with Intervals 1.2 as that version is quite old but being able to load Intervals 1.9.0 would be nice to have in Intervals 2.0.0. I would 100% punt this work until the internal redesign is done though as otherwise it'll be annoying to update while the internal design is being iterated on.

> 2. Do we want to include tests of `Base` behaviour (ie: that `union(intervalsets) != union(vectors)`)?

Yes, as long as we're testing other `Base` set functions in `comparisons.jl` we should should also include this test. We could remove the vector of intervals tests in which case we would also remove the test you mentioned. Maybe what we should so is define methods like `Base.union(::AbstractVector{Interval} = throw(MethodError)` to stop end users from mistakenly using these methods? In that case I'd be fine with removing the vector of intervals tests.

